PR #3, opened 2021-08-23
Heartbeat & Namespace Flag

zekenautilus:
## Description
- added heartbeat to config for long running services
- added namespace cli flag so that namespace in component config can be overridden at cli level

### DIFF
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -1,32 +1,35 @@
 """Summary."""
 import json
+from typing import Tuple
+from urllib.parse import urlparse
 
 import pika
 
-from typing import Tuple
-from urllib.parse import urlparse
 from src.invoker import Invoker
 from src.types import TYPE_PAYLOAD
 
 # content_type: application/json
 # {"x":5,"y":7}
 
+
 def overwrite_heartbeat(host: str, heartbeat: int) -> str:
-    u, new_param  = urlparse(host), f'heartbeat={heartbeat}'
-    found, params = False, u.query.split('&')
+    """Overwrite the heartbeat in a rabbitmq url."""
+    uri, new_param = urlparse(host), f'heartbeat={heartbeat}'
+    found, params = False, uri.query.split('&')
     for i, old_param in enumerate(params):
         if 'heartbeat' in old_param:
             found = True
             params[i] = new_param
     if not found:
         params.append(new_param)
-    return u._replace(query='&'.join(params)).geturl()
+    return uri._replace(query='&'.join(params)).geturl()
 
 
 class AmqpInvoker(Invoker):
     """Summary."""
 
-    def connect(self) -> Tuple[pika.adapters.blocking_connection.BlockingChannel, str, str]: 
+    def connect(self) -> Tuple[pika.adapters.blocking_connection.BlockingChannel, str, str]:
+        """Connect to a rabbit broker."""
         heartbeat = self._invocable.config.heartbeat
         host = overwrite_heartbeat(self._invocable.config.host, heartbeat) if heartbeat else self._invocable.config.host
         parameters = pika.URLParameters(host)
@@ -37,11 +40,11 @@ class AmqpInvoker(Invoker):
         exchange_name = self._invocable.config.exchange
         channel.queue_declare(queue=queue_name)
         channel.queue_declare(queue=queue_name_error)
+        assert isinstance(exchange_name, str)
         channel.exchange_declare(exchange_name, exchange_type='topic', passive=False, durable=True, auto_delete=False, internal=False, arguments=None)
 
-        channel.queue_bind(exchange=self._invocable.config.exchange, queue=queue_name, routing_key=str(self._invocable.config.subtopic))
+        channel.queue_bind(exchange=exchange_name, queue=queue_name, routing_key=str(self._invocable.config.subtopic))
         return channel, queue_name, queue_name_error
-
 
     def start(self) -> int:
         """Summary."""
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -3,19 +3,33 @@ import json
 
 import pika
 
+from typing import Tuple
+from urllib.parse import urlparse
 from src.invoker import Invoker
 from src.types import TYPE_PAYLOAD
 
 # content_type: application/json
 # {"x":5,"y":7}
 
+def overwrite_heartbeat(host: str, heartbeat: int) -> str:
+    u, new_param  = urlparse(host), f'heartbeat={heartbeat}'
+    found, params = False, u.query.split('&')
+    for i, old_param in enumerate(params):
+        if 'heartbeat' in old_param:
+            found = True
+            params[i] = new_param
+    if not found:
+        params.append(new_param)
+    return u._replace(query='&'.join(params)).geturl()
+
 
 class AmqpInvoker(Invoker):
     """Summary."""
 
-    def start(self) -> int:
-        """Summary."""
-        parameters = pika.URLParameters(self._invocable.config.host)
+    def connect(self) -> Tuple[pika.adapters.blocking_connection.BlockingChannel, str, str]: 
+        heartbeat = self._invocable.config.heartbeat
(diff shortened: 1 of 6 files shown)


src/config.py
```diff
@@ -58,13 +63,13 @@ def func(self) -> str:
...
     @property
-    def host(self) -> Optional[str]:
+    def host(self) -> str:
         """Summary.
 
         Returns:
             TYPE: Description
         """
-        return self._host
+        return self._host if self._host else ''
```
mattian7741, line 72: `return self._host or ''`

src/amqp_invoker.py
```diff
@@ -10,22 +12,43 @@
 # {"x":5,"y":7}
 
 
+def overwrite_heartbeat(host: str, heartbeat: int) -> str:
```
mattian7741, line 15: Can this be something more generic like "set_param"?

src/ergo_click.py
```diff
@@ -117,16 +117,18 @@ def amqp(config: Config, arg: Tuple[str]) -> int:
 
 @main.command()
 @click.argument('ref', type=click.STRING)
+@click.option('--namespace', '-n', default=None, type=click.STRING)
```
mattian7741, line 120: We talked about this and I think this is fine - certainly for now.   But I think the long term solution will probably drift away from this idea.  I could be wrong.

src/ergo_cli.py
```diff
@@ -161,9 +163,10 @@ def start(self, ref: str, *args: str) -> int:
         # use safe_load instead load
         with open(ref) as config_file:
             conf = yaml.safe_load(config_file)
-            with open(conf.get('namespace')) as namespace_file:
-                namespace = yaml.safe_load(namespace_file)
-                conf.update(namespace)
+            namespace_file_name = namespace if namespace else conf.get('namespace')
```
mattian7741, line 166: `namespace_file_name = namespace or conf.get('namespace')`

src/ergo_cli.py
```diff
@@ -147,11 +148,12 @@ def amqp(self, config: Config, *args: str) -> int:
         host.start()
         return 0
 
-    def start(self, ref: str, *args: str) -> int:
+    def start(self, ref: str, namespace: Optional[str], *args: str) -> int:
```
mattian7741, line 151: Isn't namespace available more generically in *args.   Can we avoid modifying this contract for a specific parameter?

src/config.py
```diff
@@ -58,13 +63,13 @@ def func(self) -> str:
         return self._func
 
     @property
-    def host(self) -> Optional[str]:
+    def host(self) -> str:
```
mattian7741, line 66: This seems fine to me - but is it more explicitly absent if it is null rather than empty string?  The contract is definitely stronger without the 'optional' qualifier.  I suppose since python can treat '' as null in equalities, it is ok.

src/config.py
```diff
@@ -7,6 +7,9 @@
 class Config:
     """Summary."""
 
+    # pylint: disable=too-many-instance-attributes
```
mattian7741, line 10: I agree - this is dumb

src/config.py
```diff
@@ -20,6 +23,8 @@ def __init__(self, config: Dict[str, str]):
         self._host: Optional[str] = config.get('host')
         self._exchange: Optional[str] = config.get('exchange')
         self._protocol: str = config.get('protocol', 'stack')  # http, amqp, stdio, stack
+        heartbeat = config.get('heartbeat')
+        self._heartbeat: Optional[int] = int(heartbeat) if heartbeat else None
```
mattian7741, line 27: `self._heartbeat: Optional[int] = int(config.get('heartbeat') or 0) or None`

src/config.py
```diff
@@ -83,3 +88,12 @@ def protocol(self) -> str:
             TYPE: Description
         """
         return self._protocol
+
+    @property
+    def heartbeat(self) -> Optional[int]:
```
mattian7741, line 93: Should this be part of the config?   It is an idea that shouldn't be known to the component - unless we can define it more generally than purely in the AMQP context.   Also - is it a decision that is made at the component level?  I think it's more of an environment/namespace consideration.

Another thing to consider is how config (in the code) relates to config files.   The way I think about it is that the config in code is the consolidation of all input configs (component, environment, namespace, runtime, etc.) resulting in a single running configuration represented by the Config class.  In that sense heartbeat (or the principle it represents) might make sense.

src/amqp_invoker.py
```diff
@@ -10,22 +12,43 @@
...
+        host = overwrite_heartbeat(self._invocable.config.host, heartbeat) if heartbeat else self._invocable.config.host
+        parameters = pika.URLParameters(host)
         connection = pika.BlockingConnection(parameters)
         channel = connection.channel()
         queue_name = self._invocable.config.func
         queue_name_error = f'{queue_name}_error'
         exchange_name = self._invocable.config.exchange
         channel.queue_declare(queue=queue_name)
         channel.queue_declare(queue=queue_name_error)
+        assert isinstance(exchange_name, str)
```
mattian7741, line 43: Do we need this assert?   I noticed that I didn't strongly type these variables - we should probably do that first:
Example:
  `exchange_name: str = self._invocable.config.exchange`

src/amqp_invoker.py
```diff
@@ -10,22 +12,43 @@
 # {"x":5,"y":7}
 
 
+def overwrite_heartbeat(host: str, heartbeat: int) -> str:
+    """Overwrite the heartbeat in a rabbitmq url."""
+    uri, new_param = urlparse(host), f'heartbeat={heartbeat}'
+    found, params = False, uri.query.split('&')
+    for i, old_param in enumerate(params):
```
mattian7741, line 19: Can we simplify with:
```
params.remove('heartbeat') # or if it's generalized to set_param, then old_param
params.append(new_param)
return uri._replace(query='&'.join(params)).geturl()
```

mattian7741 (commented):
Left a bunch of comments - all minor.

src/amqp_invoker.py
zekenautilus: here we're adding the param `heartbeat=3600` for example - so it wouldn't just be `params.remove('heartbeat')` but `params.remove('heartbeat=3600')`. 

But could do:
```
params = [p for p in params if param_key not in p] + [new_param]
return uri._replace(query='&'.join(params)).geturl()
```

src/amqp_invoker.py
zekenautilus: updating `Config.exchange` to be:
```
def exchange(self) -> str:
    return self._exchange or 'primary'
```

src/config.py
zekenautilus: setting `heartbeat` to 0 is a possible use case - recommended against but logically this wouldn't make it possible to set heartbeat to 0

src/config.py
zekenautilus: wasn't sure if `host` was required for _all_ types of invokers. For `amqp_invoker` surely, but `flask_http_invoker` seems to not currently use it (and not sure about invokers that would come down the line)

src/config.py
zekenautilus: I think it's a bit of a wart. Currently `pubtopic` `subtopic` and `exchange` all seem to be `amqp` specific. It would be nice if we could nest some of this by `protocol` or `namespace`.

```
parser config example
namespace:
    nautilus_qa.yaml:
        <namespace specific input here>
func: handler_parser.py:main
protocol:
    amqp:
        heartbeat: 60

```

src/config.py
mattian7741: 👍

src/amqp_invoker.py
mattian7741: Oh right.   Yeah - what you said.

src/config.py
mattian7741: Worth an independent discussion - definitely.   pubtopic and subtopic (despite their names) are actually relevant in the broader context, outside of amqp.  But this is a big topic - and not a particularly urgent one.

PR #4, opened 2021-09-02
don't publish empty results

zekenautilus:
Update to omit handling empty responses.

if an amqp func was:
```
def only_even(data):
     for num in data['nums']:
         yield num / 2 if num % 2 == 0 else yield None
```

and input was:
```
{"data": {"nums": [1, 2, 3, 4, 5, 6]}}
```

then only 3 messages will be published:
```
{"data": 1}
{"data": 2}
{"data": 3}
```

### DIFF
--- a/src/function_invocable.py
+++ b/src/function_invocable.py
@@ -59,26 +59,28 @@ class FunctionInvocable:
         self._func = arg
 
     def invoke(self, data_in: TYPE_PAYLOAD) -> Generator[TYPE_PAYLOAD, TYPE_PAYLOAD, None]:
-        """Summary.
+        """Invoke injected function.
+
+        If func is a generator, will exhaust generator, yielding each response.
+        If an exception occurs will re-raise with a stack trace.
+        Func responses will not be percolated if they return None.
 
         Args:
-            data_out (Payload): Description
-            data_in (Payload): Description
+            data_in (Payload): payload with a 'data' key. Corresponding value will be passed to injected function.
 
         Raises:
-            Exception: Description
+            Exception: caught exception re-raised with a stack trace.
 
         """
         if not self._func:
             raise Exception('Cannot execute injected function')
         try:
-            result = None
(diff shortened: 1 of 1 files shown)


mattian7741:
Is it always the case that we don't want to publish empty results?  In other words, is an empty result implicitly invalid?   I would argue that it isn't the job of the executor to make this decision.   If an "only_even" function encounters an odd number, it should throw an exception (for example) that would prevent the empty result from being published.   I haven't thought this through fully, but intuitively, this feels like overreach.

zekenautilus:
> Is it always the case that we don't want to publish empty results? In other words, is an empty result implicitly invalid? I would argue that it isn't the job of the executor to make this decision. If an "only_even" function encounters an odd number, it should throw an exception (for example) that would prevent the empty result from being published. I haven't thought this through fully, but intuitively, this feels like overreach.

@mattian7741 I think this makes more sense than throwing an exception. I'm imagining a `filter` component where maybe we want to remove topics from a feed before they get to a computationally expensive component. In such a case, by design we want to only emit topics that meet some condition or criteria - and the two of ways of doing this are:
1. not publishing topics when invoked and our input doesn't match the critera.
2. filtering on the input for the downstream, computationally expensive component. This can be alright, but in this case we may be needlessly filling the input queue w irrelevant messages.

the first seems to make more sense to me, and I'd imagine components may also want a way to opt into _not_ publishing a message. 

then to elaborate, we can either `filter` by returning `None` (which indicates ther'es nothing worth publishing) or by throwing an exception. The latter is often used for indicating an `error` - in this case it's not an error state when we're filtering, but rather expected processing. This can be confusing / unintuitive for other engineers, and also has the additional consequence of filling up our error queue.

If in our filter component, if the messages that match are condition are relatively sparse, we may end up with unnecessary processing, either in the size of our error queue relative to our output, or in the size of the input queue for the downstream computationally expensive component.

PR #5, opened 2021-09-07
python 3.7 support

zekenautilus:
## Description
- added support for python3.7 (was failing prior bc `TypedDict` was introduced in 3.8)
- added trivial unit tests to ensure import works via additional circleci python jobs


test/unit/test_util.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+from src.util import uniqueid
+
+
+class TestUtil(unittest.TestCase):
+
+    def test_util(self):
+        u1, u2 = uniqueid(), uniqueid()
+        assert u1 != u2
```
zekenautilus, line 9: super trivial - just want to verify the import in `utils.py` works.

src/util.py
```diff
@@ -61,15 +66,15 @@ def get_stack() -> List[FrameType]:
...
+        trcbk = trcbk.tb_next  # pragma: no cover
     stack: List[FrameType] = []
     while frm:
         stack.append(frm)
         frm = frm.f_back
     return stack
 
 
-def print_exc_plus() -> str:
+def print_exc_plus() -> str:  # pragma: no cover
```
zekenautilus, line 77: codecov was failing on these lines bc this file wasn't test before and thus coverage on it was "100%" - figured it might be fine to ignore these specific lines for now

ahuman-bean: I think aiming for 100% is ambitious, because sometimes tests can be trivialized in achieving that coverage. You can configure the threshold w/ a `codecov.yml` file so there's some leeway, e.g. drop of <1% in coverage still passes.

ahuman-bean:
A bit of a nitpick: https://github.com/mattian7741/ergo/blob/master/setup.py#L97-L100. Are we ready to update this and claim to be compliant for all of these versions?

zekenautilus:
> A bit of a nitpick: https://github.com/mattian7741/ergo/blob/master/setup.py#L97-L100. Are we ready to update this and claim to be compliant for all of these versions?

am a little confused - the setup py classifiers imply 3, 3.4 -> 3.6 support, but the tests show 3.6 -> 3.9 support... curious what happens if I update to add more python version tests. The only reason why I'm hesitant to update is bc these tests while they claim high coverage, don't really test the ergo cli, or ergo commands, just that this package can be imported and installed really.

PR #7, opened 2021-09-16
Flask bugfix

zachschubert:
This refactors `ERGO_CLI.http` to accept a path to a function instead of a path to a config, so that the example in the README works again. To test:
```
$ ergo http test/integration/test_http.py:product
```
It also adds a few integration tests for that endpoint, mostly taking different approaches to that README example. Currently the only test that passes is `test_payload`, because the `FunctionInvocable` seems to [assume](https://github.com/mattian7741/ergo/blob/2eb2a2920fb2d2abc4cad87ac9e415825dcd6cb9/src/function_invocable.py#L81) that ergo-wrapped functions always have a single parameter, and that requests always supply it in a mapping where `data` is the key. I don't have a firm opinion on which set of these calling patterns we want to support, although I'm not sure what the benefit is for requiring that the arguments be wrapped in a `data` key. 

Finally, this refactors `FlaskHttpInvoker` to serialize output with a custom JSON encoder. This is required for the integration tests to [parse](https://github.com/mattian7741/ergo/blob/4dd4fb0c26b96abcbf7d72fee7865860f5898766/test/integration/test_http.py#L42) the response bodies, even for subjects like `product` that return a simple scalar.


test/integration/test_http.py
```diff
@@ -0,0 +1,95 @@
...
+
+class TestProduct(HTTPTest):
+    target_func = product
+
+    def test(self):
+        """tests the example function from the ergo README"""
+        resp = self.session.get("http://localhost?4&5")
+        self.assertEqual(resp.status_code, 200)
+        body = resp.json()
+        self.assertEqual(body[0]["data"], 20)
```
zachschubert, line 43: This response body format is also a departure from the README example. Which format do we prefer?

zachschubert: I guess my real question is how is that `log` data ergo returns meant to be used, and why are it and `data` wrapped in a list?

zachschubert:
brb this isn't ready

PR #10, opened 2021-09-23
Various fixes

zachschubert:
1. Refactor FunctionInvocable.invoke to take an arbitrary dict (instead of requiring one with a `data` key) and splat it into the injected function. This is necessary to let ergo wrap functions that e.g. don't have any params at all, like [this case](https://github.com/mattian7741/ergo/blob/be459ee5e39457d20e95fc7d7642f53a2f1a2688/test/integration/test_http.py#L37-L61). It's also necessary to support [components](https://github.com/mattian7741/ergo/blob/dea9bbb58cdb76bd17d29a5930ee145fed404b19/test/integration/test_http.py#L40-L61) that don't take any args at all.
2. Refactor `ergo http` to accept either a path to a manifest file or a function string, as in the README example. 
3. Add some integration tests for `ergo start amqp`, `ergo start http` and `ergo http`.


README.md
```diff
@@ -25,6 +25,6 @@ $ ergo http math.py:product
 then to make a request against this service
 
 ```
-$ curl -i "http://localhost?4&5"
+$ curl -i "http://localhost?x=4&y=5"
```
zachschubert, line 28: We could support the original, but it would be somewhat tricky. Is it worth it?

src/amqp_invoker.py
```diff
@@ -19,6 +17,11 @@ def set_param(host: str, param_key: str, param_val: str) -> str:
     return uri._replace(query='&'.join(params)).geturl()
 
 
+def declare_topic_exchange(channel, exchange_name):
+    channel.exchange_declare(exchange_name, exchange_type='topic', passive=False, durable=True, auto_delete=False,
+                             internal=False, arguments=None)
```
zachschubert, line 22: I pulled this out because the test needs to asynchronously declare this exchange with all of the same params, or else it'll break inside rabbitmq.

test/integration/test_amqp.py
```diff
@@ -0,0 +1,100 @@
...
+from src.topic import PubTopic
+from src.amqp_invoker import declare_topic_exchange
+from test.integration.utils import with_ergo
+
+
+AMQP_HOST = "amqp://guest:guest@localhost:5672/%2F"
+
+
+@pytest.fixture(scope="session")
+def rabbitmq():
```
zachschubert, line 17: IMO there's no need to teardown this container automatically, but I'm open to reconsidering.

src/function_invocable.py
```diff
@@ -75,10 +75,9 @@ def invoke(self, data_in: TYPE_PAYLOAD) -> Generator[TYPE_PAYLOAD, TYPE_PAYLOAD,
         if not self._func:
             raise Exception('Cannot execute injected function')
         try:
-            if inspect.isgeneratorfunction(self._func):
-                result_exp = self._func(data_in['data'])
-            else:
-                result_exp = (r for r in [self._func(data_in['data'])])
+            result_exp = self._func(**data_in)
```
ahuman-bean, line 78: are the keys for `data_out` and `data_in` consistent w/o explicitly needing a top-level `data` key?

on line 84 below we could probably un-nest the result if we're saying components should consume both 'data' and routing/debug-level information

src/ergo_click.py
```diff
@@ -82,20 +82,20 @@ def use(name: str) -> int:
...
-@click.argument('ref', type=click.STRING)
+@click.argument('target', type=click.STRING)
 @click.argument('arg', nargs=-1)
-def http(config: Config, arg: Tuple[str]) -> int:
+def http(target: str, arg: Tuple[str]) -> int:
     """Summary.
 
     Args:
-        config (str): Description
+        target (str): Path to a configuration file or to a function.
```
ahuman-bean, line 91: This isn't true since we are using a specific pattern string

src/flask_http_invoker.py
```diff
@@ -36,8 +38,16 @@ def handler() -> str:  # type: ignore
...
-            return str(data_out)
+            return json.dumps(data_out, cls=ErgoEncoder)
 
         app.run(host='0.0.0.0', port=self._port)
         return 0
+
+
+class ErgoEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, datetime.datetime) or isinstance(obj, datetime.date):
```
ahuman-bean, line 49: @mattian7741 @zachschubert I like having reasonable defaults, but for something like `datetime` we would be taking a stance on either unix timestamp or iso string. the latter is the lowest lift for consumers. I think we are somewhat walking a fine line here unless `ergo` is more transparent (maybe through documentation) about how certain types will be serialized by default

ahuman-bean:
@zachschubert https://circleci.com/docs/2.0/building-docker-images/ I believe this should resolve the connection issue when testing docker.

README.md
```diff
@@ -25,6 +25,6 @@ $ ergo http math.py:product
 then to make a request against this service
 
 ```
-$ curl -i "http://localhost?4&5"
-{"result": 20.0}
+$ curl -i "http://localhost?x=4&y=5"
+[20.0]
```
ahuman-bean, line 29: We should prefer a root object instead of an array

zachschubert: ⚡ i agree

README.md
mattian7741: It's not important.   I think during initial development both were supported implicitly, but I don't know what the exact circumstances were.

src/ergo_cli.py
```diff
@@ -117,17 +117,21 @@ def init(self, name: str) -> int:
         self.use(name)
         return 0
 
-    def http(self, config: Config, *args: str) -> int:
+    def http(self, target, *args: str) -> int:
```
mattian7741, line 120: Can you provide a type for target `target: str`

Also - thanks for picking this error up - this is another example of behavior that was forgotten, so this is a good catch.   The only feedback here is that I don't think the http directive should support `config`.  I think it will be cleaner and more valid only to support target in this case.  What do you think?

src/amqp_invoker.py
```diff
@@ -53,11 +50,15 @@ def handler(channel, method, properties, body) -> None:  # type: ignore
...
             """
             data_in: TYPE_PAYLOAD = dict(json.loads(body.decode('utf-8')))
-            data_in['key'] = str(self._invocable.config.subtopic)
             try:
-                for data_out in self._invocable.invoke(data_in):
-                    data_out['key'] = str(self._invocable.config.pubtopic)
-                    channel.basic_publish(exchange=self._invocable.config.exchange, routing_key=str(self._invocable.config.pubtopic), body=json.dumps(data_out))
+                for data_out in self._invocable.invoke(data_in["data"]):
+                    payload = {
+                        "data": data_out,
```
mattian7741, line 56: I am still thinking about this - because while I know intuitively (and from vague memory of my own pitfalls) that there is no logical difference between http and amqp as it relates to this issue, I am having trouble articulating the logic just yet.   For now - let's move with this, and if I can get my thoughts together we can come back to it.

PR #11, opened 2021-09-28
build and test with docker-compose

zachschubert:
This refactors the circleci test job to run tests inside a custom `ergo` docker container. The motivation is to support https://github.com/mattian7741/ergo/pull/10, which needs to run tests in an environment that has `rabbitmq-server` installed. 

I initially tried running rabbitmq in a separate docker container via circleci's [remote docker environment](https://circleci.com/docs/2.0/building-docker-images/), but that container's broker wasn't reachable at `amqp://guest:guest@localhost:5672`. I figured it would be better to solve it the way I did here, so that the test doesn't have to do environment-dependent service discovery.


.circleci/config.yml
```diff
@@ -1,64 +1,22 @@
...
+      python_version:
         type: string
-        default: latest
-    docker:
-      - image: circleci/python:<< parameters.version >>
+    environment:
+      ERGO_PYTHON_VERSION: << parameters.python_version >> 
     steps:
       - checkout
-      - restore_cache:
```
ahuman-bean, line 37: Is there any reason why we are removing dependency caching? These tend to speed up builds quite a bit

ahuman-bean: 🤔 

Ok I see, do you know if docker layer caching is available when using docker / docker-compose separate from circleci?

ahuman-bean: I'm curious if `machine: true` uses `docker_layer_caching: true` by default.

ahuman-bean (approved):
Nice!

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+FROM python:${ERGO_PYTHON_VERSION}-slim as ergo-python
+
+RUN apt-get update && apt-get install -y make
+
+COPY requirements.txt /tmp/requirements.txt
+RUN pip install -r /tmp/requirements.txt && rm /tmp/requirements.txt
+
+COPY dev-requirements.txt /tmp/dev-requirements.txt
+RUN pip install -r /tmp/dev-requirements.txt && rm /tmp/dev-requirements.txt
```
zachschubert, line 11: FYI I'm doing this instead of using `make dev` because we get much better docker image layer caching this way.

.circleci/config.yml
zachschubert: Screen Shot 2021-09-28 at 4 14 49 PM

:(

zachschubert:
I figured out how to do https://github.com/mattian7741/ergo/pull/10 the easy way with a circlecli remote docker engine

PR #18, opened 2021-10-08
start rabbitmq broker before launching pytest

zachschubert:
I added the 30s per-test timeout to the `make test` command because amqp integration tests can potentially hang forever waiting for messages. But the first amqp test run can take longer than 30s if it has to start a rabbitmq docker container. This is causing some [flakiness](https://app.circleci.com/pipelines/github/mattian7741/ergo/366/workflows/fed49bdc-df6c-4249-8e89-c8ed8409c8a7/jobs/1122) on circleci. IMO the best solution is to ensure the broker is ready before launching pytest, so that we can keep the 30s timeout (which is probably actually way too long).


test/integration/start_rabbitmq_broker.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Start a rabbitmq server if none is running, and then wait for the broker to finish booting.
+    """
+    docker_client = docker.from_env()
+    if not docker_client.containers.list(filters={"name": "rabbitmq"}):
+        docker_client.containers.run(
+            name="rabbitmq",
+            image="rabbitmq:3.8.16-management-alpine",
+            ports={5672: 5672, 15672: 15672},
+            detach=True,
```
ahuman-bean, line 20: would be nice to be able to pull this out from an existing `docker-compose.yml` for dev'ing against this project

PR #38, opened 2021-10-29
extend amqp invoker to support rpc calls

zachschubert:
This adds `ergo_python.rpc_client.ErgoRPCClient`, a utility for making synchronous rpc calls to be handled by one or multiple amqp components.

To implement this, I introduced Graph as a new parent class of Config. In principle, I'm imagining Graph as an abstract application that takes requests at one topic and optionally publishes results to another. A graph would then be concretely implemented by one or more components. The practical purpose is that Graph contains all of the information necessary for a client to make an ergo RPC call. I'm not sure this is the best name for this concept. Other alternatives I'm thinking about are Application, Service, or Process.


src/amqp_invoker.py
```diff
@@ -142,12 +143,23 @@ async def on_message(self, message: aio_pika.IncomingMessage, channel: aio_pika.
...
             except Exception as err:  # pylint: disable=broad-except
                 data_in['error'] = str(err)
                 # TODO(ahuman-bean): cleaner error messages
-                await exchange.publish(message=aio_pika.Message(body=json.dumps(data_in).encode()), routing_key=f'{self.queue_name}_error')
+                out_message = aio_pika.Message(body=json.dumps(data_in).encode(),
+                                               correlation_id=message.correlation_id)
+                routing_key = self._invocable.config.pubtopic
+                if message.reply_to:
+                    routing_key = routing_key.extend(message.reply_to)
+                await exchange.publish(message=out_message, routing_key=str(routing_key))
```
ahuman-bean, line 162: nit: str cast also no longer needed now

src/amqp_invoker.py
```diff
@@ -142,12 +143,23 @@ async def on_message(self, message: aio_pika.IncomingMessage, channel: aio_pika.
...
+                            routing_key = routing_key.extend(message.reply_to)
+                        await exchange.publish(message=out_message, routing_key=str(routing_key))
 
             except Exception as err:  # pylint: disable=broad-except
                 data_in['error'] = str(err)
                 # TODO(ahuman-bean): cleaner error messages
-                await exchange.publish(message=aio_pika.Message(body=json.dumps(data_in).encode()), routing_key=f'{self.queue_name}_error')
+                out_message = aio_pika.Message(body=json.dumps(data_in).encode(),
+                                               correlation_id=message.correlation_id)
+                routing_key = self._invocable.config.pubtopic
```
ahuman-bean, line 159: this used to route to the error queue. I guess we are going to propagate errors downstream then?

src/topic.py
```diff
@@ -13,9 +13,28 @@ def __init__(self, topic_str: Optional[str]):
...
+        >>> extended = topic.extend("c")
+        >>> str(topic)
+        'a.b'
+        >>> str(extended)
+        'a.b.c'
+
+        """
+        if self._raw:
+            topic_str = ".".join([self._raw, topic_str])
+        return type(self)(topic_str)
```
ahuman-bean, line 34: can do something like:

```python
# Note the quotations around the identifier
def extend(self, topic_str: str) -> "Topic":
```

and don't we want to update the underlying value?

so maybe:
```python
        if self._raw:
            self._raw = ".".join([self._raw, topic_str])
        return self
```

or if immutable and constructing a new object and returning, this could also be a `staticmethod` taking in two params

PR #39, opened 2021-11-02
fix(amqp_invoker): better catch-all amqp exception retries

ahuman-bean:
Now catching base `AMQPError`, since `AMQPConnectionError` was a bit too specific to the `connection` handle.

Also the occasional `ChannelInvalidStateError` inherits from `RuntimeError`.

### DIFF
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -57,7 +57,7 @@ class AmqpInvoker(Invoker):
 
         return 0
 
-    @retry(aio_pika.exceptions.AMQPConnectionError, delay=1, jitter=(1, 3), backoff=2)
+    @retry((aio_pika.exceptions.AMQPError, aio_pika.exceptions.ChannelInvalidStateError), delay=1, jitter=(1, 3), backoff=2)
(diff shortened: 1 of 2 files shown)


src/amqp_invoker.py
```diff
@@ -73,35 +73,32 @@ async def run(self, loop: asyncio.AbstractEventLoop) -> aio_pika.RobustConnectio
...
-                    futures = [
-                        self.consume(channel, queue, self.on_message),
-                        # TODO(ahuman-bean): implement error handling
-                        # self.consume(channel, queue_error, self.on_error)
-                    ]
-
-                    # Gathers all workers as futures and runs them concurrently on the underlying execution context
-                    await asyncio.gather(*futures)
-                except aio_pika.exceptions.ConnectionClosed:
-                    pass
```
ahuman-bean, line 104: removed this for now only because I'm not sure how we will be logging at this level (so no stance on internal logging for now)

ahuman-bean: also it was a bit too specific of an exception that was set from an early iteration.

zachschubert (approved):

PR #41, opened 2021-11-10
refactor(amqp): Channel pooling for consuming/publishing

ahuman-bean:
**Context**
Consistently reproduced `InvalidChannelStateError` thrown when `AsyncQueueIteratorContextManager` exits too early for long running consume/publish tasks, >30mins and test with >4hr long tasks. It's possible that although the `connection` was kept alive, there was an underlying issue with the `channel`/`queue` handles that were no longer valid to be reused by the publisher-worker after completing its work in 4hrs.

**Changes**

* Consumer and publisher within event loop attempt to reacquire channel handles prior to each operation
* Avoid using async queue iterator for long running work
https://github.com/mosquito/aio-pika/blob/master/aio_pika/queue.py#L358-L379
* Also removed stubbed `on_error` callback + generally simplified code

**Notes**

* python 3.6 breaks unfortunately bc of aio-pika's poor support, but for what it's worth, asyncio in general is very new to 3.6 and prior anyways
* more granular testing to track which channels/threads in pool/executor are being used https://github.com/mattian7741/ergo/issues/36
* integration test case - existing cases hold for typical use, question is if we want to put in an optional test case where we sleep for >45min to cover the edge case the PR addresses, simply because it would take so long and hang up CI/CD

### DIFF
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -57,7 +57,7 @@ class AmqpInvoker(Invoker):
 
         return 0
 
-    @retry(aio_pika.exceptions.AMQPConnectionError, jitter=(1, 3), backoff=2)
+    @retry((aio_pika.exceptions.AMQPError, aio_pika.exceptions.ChannelInvalidStateError), delay=0.5, jitter=(1, 3), backoff=2)
     async def run(self, loop: asyncio.AbstractEventLoop) -> aio_pika.RobustConnection:
         connection = await aio_pika.connect_robust(url=self.url, loop=loop)
 
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -116,9 +116,8 @@ class AmqpInvoker(Invoker):
             queue: Readable message queue
             callback: Awaitable message handler
         """
-        async with queue.iterator() as consumed:
-            async for message in consumed:
-                await callback(message, channel)
+        async for message in queue:
+            await callback(message, channel)
 
     async def on_message(self, message: aio_pika.IncomingMessage, channel: aio_pika.RobustChannel) -> None:
         """
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -137,7 +137,7 @@ class AmqpInvoker(Invoker):
         Parameters:
             channel_pool: Pool of valid `Channel` handles
             message: Message to be published
-            routing_key: Exchange-level discriminator
+            routing_key: Exchange-level routing discriminator
         """
         async with channel_pool.acquire() as channel:
             exchange = await channel.get_exchange(name=self.exchange_name, ensure=False)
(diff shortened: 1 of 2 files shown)


src/amqp_invoker.py
```diff
@@ -57,104 +57,88 @@ def start(self) -> int:
 
         return 0
 
-    @retry((aio_pika.exceptions.AMQPError, aio_pika.exceptions.ChannelInvalidStateError), delay=1, jitter=(1, 3), backoff=2)
+    @retry((aio_pika.exceptions.AMQPError, aio_pika.exceptions.ChannelInvalidStateError), delay=0.5, jitter=(1, 3), backoff=2)
```
zekenautilus, line 60: any specific reasons for the update in delay?

ahuman-bean: 0.5s vs. 1s on the first reconnect. just being more aggressive at first, backoff will handle the rest

src/amqp_invoker.py
```diff
@@ -57,104 +57,91 @@ def start(self) -> int:
...
+                name=self.exchange_name,
+                type=aio_pika.ExchangeType.TOPIC,
+                passive=False,
+                durable=True,
+                auto_delete=False,
+                internal=False,
+                arguments=None
+            )
+
+        # Consume-Publish loop
```
zekenautilus, line 89: is it possible the Master-Worker pattern could be helpful here? https://aio-pika.readthedocs.io/en/latest/patterns.html#master-worker

ahuman-bean: No. `Master` refers to the upstream component. `Worker` refers to this downstream consumer component. We are trying to achieve an internal consumer -> do work -> publish workflow.

src/amqp_invoker.py
```diff
@@ -57,104 +57,91 @@ def start(self) -> int:
...
-                    async for data_out in generator:
-                        await exchange.publish(message=aio_pika.Message(body=json.dumps(data_out).encode()), routing_key=str(self._invocable.config.pubtopic))
+            await queue.bind(exchange=exchange, routing_key=str(self._invocable.config.subtopic))
+            await queue_error.bind(exchange=exchange, routing_key=f'{self.queue_name}_error')
 
-            except Exception as err:  # pylint: disable=broad-except
-                data_in['error'] = str(err)
-                # TODO(ahuman-bean): cleaner error messages
-                await exchange.publish(message=aio_pika.Message(body=json.dumps(data_in).encode()), routing_key=f'{self.queue_name}_error')
+            async for message in queue:
```
zekenautilus, line 129: is this chill with multiple queue consumers? I.e. 2 ergo components consuming from the same queue? I assume ye since it's `async` but just checking...

src/amqp_invoker.py
```diff
@@ -57,104 +57,91 @@ def start(self) -> int:
...
-            message: Message object with convenience methods for acknowledgement
-            channel: Connection handle and state
+            channel_pool: Pool of valid `Channel` handles
+            message: Message to be published
+            routing_key: Exchange-level discriminator
         """
-        pass
+        async with channel_pool.acquire() as channel:
+            exchange = await channel.get_exchange(name=self.exchange_name, ensure=False)
+            await exchange.publish(message, routing_key)
```
zekenautilus, line 144: this is the crux of the change right? on publish we reacquire the channel? in case the channel got into a wrong state?

zekenautilus (approved):
lgtm - excluding the python3.6 and python3.7 tests failing, sure there's a workaround for that

src/amqp_invoker.py
ahuman-bean: Everything inside of `async def run` is internally synchronous, because we've not scheduled any coroutine via a variant of `loop.run...`

PR #42, opened 2021-11-12
fix(amqp): Ack message before processing to satisfy broker consumer timeout

ahuman-bean:


zekenautilus (approved):
personally think it makes sense to ack early so long as we have some permanent record of the messages that are passing through our queues. Unfortunately, we don't currently have that, but maybe we move forward 'at risk' w this change, assuming that the permanent solution comes rather soon.

ahuman-bean:
> personally think it makes sense to ack early so long as we have some permanent record of the messages that are passing through our queues. Unfortunately, we don't currently have that, but maybe we move forward 'at risk' w this change, assuming that the permanent solution comes rather soon.

haven't really discussed it w/ the larger team yet. will hold off since this could probably be an opt-in sort of change

src/amqp_invoker.py
```diff
@@ -127,9 +127,15 @@ async def consume(self, channel_pool: aio_pika.pool.Pool[aio_pika.RobustChannel]
...
-                async with message.process():
-                    yield dict(json.loads(message.body.decode('utf-8')))
+            async with queue.iterator() as iter:
+                async for message in iter:
+                    data_in: TYPE_PAYLOAD = dict(json.loads(message.body.decode('utf-8')))
+                    # Subsequent processing work could still fail, but message delivery is considered to be successful since the `message` was not structurally malformed
+                    # Therefore, acknowledgement is sent upon successful receipt and deserialization of `message`
+                    # This is performed prior to processing due to the `consumer_timeout` (default 30 minutes) on the broker side
+                    # See: https://www.rabbitmq.com/consumers.html#acknowledgement-timeout
+                    await message.ack()
```
zachschubert, line 137: Why deserialize before acking? If one consumer fails to deserialize, isn't everybody else guaranteed to?

ahuman-bean: We may want to leverage the error queue for something like this as well

PR #43, opened 2021-11-22
fix(amqp): replace '.' topic separator in err queue binding w/ '_'

ahuman-bean:
## Description

**Changes**

#### `.` is used as a topic separator: https://github.com/nautiluslabsco/ergo/blob/master/src/topic.py#L18

So currently, it is impossible to bind to any error queue containing `.` in its name.

#### Wrap/nest error payload/metadata in another `"data"` key to expose to downstream consumers

https://github.com/nautiluslabsco/ergo/pull/43/files#diff-408af79b4dbd03ba8e104983e4c47cca06fe51d3a90da05e79d865a0f22218a3R101-R102

## Example

Given a queue binding: `product.py:compute`, a corresponding `product.py:compute_error` queue / routing_key is also generated.
Attempting to bind a new queue to routing key `product.py:compute_error` would instead result in a binding of `#.py:compute_error.#.product.#`

## Notes

This would be a breaking change.

### DIFF
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -96,9 +96,10 @@ class AmqpInvoker(Invoker):
                         await self.publish(channel_pool, message, routing_key=routing_key)
 
                 except Exception as err:  # pylint: disable=broad-except
-                    data_in['error'] = str(err)
                     # TODO(ahuman-bean): cleaner error messages
-                    message = aio_pika.Message(body=json.dumps(data_in).encode())
+                    data_in['error'] = str(err)
+                    # NOTE(ahuman-bean): have to nest metadata so error consumers can actually have access to them
+                    message = aio_pika.Message(body=json.dumps({"data": data_in}).encode())
                     routing_key = f'{self.queue_name.replace(".", "_")}_error'
                     await self.publish(channel_pool, message, routing_key)
 
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -99,7 +99,7 @@ class AmqpInvoker(Invoker):
                     data_in['error'] = str(err)
                     # TODO(ahuman-bean): cleaner error messages
                     message = aio_pika.Message(body=json.dumps(data_in).encode())
-                    routing_key = f'{self.queue_name}_error'
+                    routing_key = f'{self.queue_name.replace(".", "_")}_error'
                     await self.publish(channel_pool, message, routing_key)
 
         return connection
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -124,7 +124,7 @@ class AmqpInvoker(Invoker):
             queue_error = await channel.declare_queue(name=f'{self.queue_name}_error')
 
             await queue.bind(exchange=exchange, routing_key=str(self._invocable.config.subtopic))
-            await queue_error.bind(exchange=exchange, routing_key=f'{self.queue_name}_error')
+            await queue_error.bind(exchange=exchange, routing_key=f'{self.queue_name.replace(".", "_")}_error')
 
             async for message in queue:
                 async with message.process():


src/amqp_invoker.py
```diff
@@ -96,10 +96,11 @@ async def get_channel() -> aio_pika.Channel:
...
 
                 except Exception as err:  # pylint: disable=broad-except
-                    data_in['error'] = str(err)
                     # TODO(ahuman-bean): cleaner error messages
-                    message = aio_pika.Message(body=json.dumps(data_in).encode())
-                    routing_key = f'{self.queue_name}_error'
+                    data_in['error'] = str(err)
+                    # NOTE(ahuman-bean): have to nest metadata so error consumers can actually have access to them
+                    message = aio_pika.Message(body=json.dumps({"data": data_in}).encode())
+                    routing_key = f'{self.queue_name.replace(".", "_")}_error'
```
zekenautilus, line 103: if the queue name is being replaced like this twice it might make more sense to store as a member of self?
`self._queue_error_name = f'{self.queue_name.replace(".", "_")}_error`?

zekenautilus:
not super sure if this is the best step forward... A couple thoughts:
- is it possible for us to update our subscribing, such that we may subscribe to something like:
`#:compute_error`? Maybe a more standard way of subscribing to error queues, rather than an update to how we create error queues? The `:` should give us some flexibility as a special key. Maybe when subscribing and we see a `:` it indicates we should not treat `.` characters as topic separators, but rather as normal strings, and don't perform any processing
- in rabbit `.` is used to separate 'words' in topic bindings - makes it feel awkward that we have `.py` in our queue names considering `.py` is part of our function name, rather than some meaningful 'word' in our topic. 
  - Something simple we can do is stop appending `.py` to our function entry points - i.e. `product.py` is just `product`
  - maybe we update the generated queue names? Instead of a queue such as `product.py:compute_error` we remove `.py` and create a queue named `product:compute_error` and `product:compute`? The former's binding then doesn't have any `.` in it

src/amqp_invoker.py
mattian7741: I'm not sure why any of this is causing an issue.   The queue name shouldn't affect routing in any way - the only time the queue name is relevant in routing is in a direct exchange.   I would need to understand how we are experiencing issues to provide better feedback here, but I think while the substitution of "_" for "." might make the problem go away, I don't think it should be necessary.   Also - loosely related - is the secondary per component-type exchange which we need for retries, rather than permanent errors.  Might be easier to resolve this in a quick discussion.

zekenautilus: currently we're publishing error messages with a routing key set to `f'{self.queue_name}_error'`. Can see that here:
https://github.com/nautiluslabsco/ergo/blob/master/src/amqp_invoker.py#L101-L106

So if someone wants to subscribe to these error messages then they would need to set their routing key to something similar. For example if:
```
config.yaml
---
func: example.py:product
```

when we run into an exception, we're going to be publishing with the routing key:
`example.py:product_error`


It's currently impossible to subscribe to these error messages using that routing key in the config.

zekenautilus: can see an alternative solution here: https://github.com/nautiluslabsco/ergo/pull/46

mattian7741: A few things here:
If we want to support subscriptions to error messages (and we might not want to) then I think we should do it differently - something along the lines of <original_topic>.error

There are a couple of other points here.  The first is that the routing key (name of the queue) using the default exchange (ie. directly to the consumer queue) should not be modified by our higher level topic processing (ie. sorting, wildcarding, etc.).    But I just reread my previous comment, which is incorrect; we don't want to use the default exchange because we want to preserve the original routing key in the message.   The only way to do this (as far as I know) is to introduce an error exchange specific to the consumer.    So at the time that we automatically generate the queue and the bindings for the consumer, we should also create a permanent error queue, and two additional exchanges: One for direct routing into the consumer, and the other for routing into the perm. error queue.   The original topic should be used in all cases - although there is nothing stopping us from augmenting the topic (ie. <original_topic>.error).   The new exchanges bind with "#" to their respective queues.

Diagram attache
Ergo Error Handling
d.

zekenautilus: The problem with publishing w `<original_published_topic>.error` is that there might be some downstream consumers that are interested in `<original_published_topic>` but _not_ with `<original_published_topic>.error`. 

For example if a downstream consumer wants to subscribe to: `fox.orange` it's actual binding at runtime will be:
`#.fox.#.orange.#`. Which will match against `error.fox.orange` or `fox.orange.error`.

The purpose for consuming errors here isn't so much for _retrying_ . Imagine we have something like:

```
scheduler -> expensive_processor -> successful email alerter
                                                            -> failure email alerter
```

in this case `expensive_processor` might work, might not (due to external factors... like dependencies on third party apis). In case of success, we want to process its output a little further and then email a client. In case of failure however, we want to email stakeholders (the programmer who designed the component, or the pm that invoked it w specific parameters) that those parameters didn't work for _x_ reason.

mattian7741: It's a different exchange.   So there won't be any downstream consumers.

mattian7741: The key point here is that there are only two resultant conditions with a consumer:
1. It does its work and succeeds and explicitly publishes some outbound message
2. It fails and retries and permanent errors occur within the scope of that component only - not on the public network.

If a component fails with errors that are meaningful to the public network, then the component should publish explicit messages to that effect.

Issue #23, opened 2021-10-11
Distinguish Traceback from Exception


zekenautilus:
## Context
Currently the `error` field of an ergo error is a massive traceback leading up to the exception itself. This is useful information, but maybe there's a better way of doing this so as to not clutter the message? Particularly relevant when trying to debug multiple error messages at a glance.

## Description
Solutions can be:
1. move `error` field to `traceback` and re-introduce the `error` key as _just_ the exception
2. only output traceback at the logging level
3. introduce additional keys similar to other logging mechanisms such as `file` and `lineno`, removing `traceback` altogether

PR #44, opened 2021-11-24
separate traceback from exception

zekenautilus:
## Description

Issue: https://github.com/nautiluslabsco/ergo/issues/23

Separating `error` from `traceback` so that we can at a glance debug several messages in the rabbitmq console. This leverages exception chaining: https://docs.python.org/3/tutorial/errors.html#exception-chaining to grab the originating exception and the traceback information from the same object.

## Visuals

function code that leads to index error:
Screen Shot 2021-11-24 at 11 05 35 AM

input message:
Screen Shot 2021-11-24 at 11 04 16 AM

error message:
Screen Shot 2021-11-24 at 11 05 17 AM

note that `error` now says the more human readable:
`TypeError: 'int' object is not subscriptable`

## Thoughts
There are some annoying linting errors that are popping up... think it would be nice if linting wasn't kicked off pre-commit but rather as a blocker to merging. Maybe linting should automatically clean up the code rather than whine about it as well?

### DIFF
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -26,7 +26,7 @@ def set_param(host: str, param_key: str, param_val: str) -> str:
     return uri._replace(query='&'.join(params)).geturl()
 
 
-def make_error_output(err: Exception) -> Dict:
+def make_error_output(err: Exception) -> Dict[str, str]:
     """Make a more digestable error output."""
     orig = err.__context__
     err_output = {
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -1,7 +1,7 @@
 """Summary."""
 import asyncio
 import json
-from typing import Iterable
+from typing import Dict, Iterable
 from urllib.parse import urlparse
 
 import aio_pika
@@ -11,6 +11,7 @@ from retry import retry
 from src.function_invocable import FunctionInvocable
 from src.invoker import Invoker
 from src.types import TYPE_PAYLOAD
+from src.util import extract_from_stack
 
 # content_type: application/json
 # {"x":5,"y":7}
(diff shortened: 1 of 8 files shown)


src/version.py
```diff
@@ -7,7 +7,7 @@
 import subprocess
 import sys
 
-VERSION = '0.5.5-alpha'
+VERSION = '0.5.6-alpha'
```
ahuman-bean, line 10: I was about to say that this could technically be `0.6.0-alpha`, except that even if a component were able to subscribe to the queue, they currently would only see `data` and neither `error` nor `traceback` (so would not be a breaking change.

ahuman-bean (approved):
💯

src/amqp_invoker.py
```diff
@@ -96,8 +86,9 @@ async def get_channel() -> aio_pika.Channel:
                         await self.publish(channel_pool, message, routing_key=routing_key)
 
                 except Exception as err:  # pylint: disable=broad-except
-                    data_in['error'] = str(err)
-                    # TODO(ahuman-bean): cleaner error messages
+                    orig = err.__context__
+                    data_in['error'] = f'{type(orig).__name__}: {str(orig)}'
```
ahuman-bean, line 90: is there a way to make this structured? so instead of `data_in['error']: str`, maybe could use a structure instead

```
type Error =
   type: Type
   message: String
   line: Number
```

e.g.:

`{"error": { "type": "TypeError", "message": "int object is not subscriptable", "line": 114 } }`

zekenautilus: maybe something like 
image
?

zekenautilus: also going to look into: https://docs.python.org/3/library/traceback.html#tracebackexception-objects

ahuman-bean: was mostly just curious if we could get file/line info of the primary exception/error in there as well.

but LGTM regardless

zekenautilus: TracebackException _would_ be useful if it didn't just set these members when _syntax errors_ occurred:

image

zekenautilus: @ahuman-bean thoughts on something like this?

image

some steps forward:
- can regex extract lineno, file, and such
- can add another top level field (`stack`)?
- maybe this is fine

ahuman-bean: yep next step imo would be to regex extract. "stack" probably won't be necessary in that case, since it refers to "error"

ahuman-bean: 😞 

for syntax errors only

zekenautilus: most recent update:

image

Issue #22, opened 2021-10-11
AMQP Dynamic Routing


zekenautilus:
## Context
Currently we can only publish against a singular binding specific in the config file. Components don't have any mechanism to dynamically generate bindings based on their input (or any application logic)

## Description
Add an interface, or hook for components to modify the binding they are publishing against. Maybe for example:
```
def bad_example(key):
     if key == "a":
        # publish against #.A.# binding
     elif key == "b":
        # publish against #.B.# binding
```

## Initial Proposed Solution

An idea that might be worth exploring:
we've worked before w libraries / applications that provide different levels of information to the consumer based on the function signature the consumer puts forth. Maybe we could do the same??

For example if we have:

```
product.yaml
---
subtopic: nums
pubtopic: product

handler.py
---
def product(data):
    yield data['x'] * data['y']
```

this will publish against `#.product.#` per the configuration. But maybe we can also support something like:
```
product.yaml
---
subtopic: nums
pubtopic: product

handler.py
---
def product(data, context):
    if 'do_sum_instead' in data:
        context['pubtopic'] = '#.sum.#'
        yield data['x'] + data['y'], context
    yield data['x'] * data['y'], context
```
that will publish against `#.sum.#` if a certain flag is set in the payload, and against `#.prod.#` otherwise!

## Rough Implementation
Implementation shouldn't be too hard. Maybe something like:
```
if function_signature == func(data, context):
    output_data, output_context = func(payload['data'], populate_context())
    output_key = output_context.get('pubtopic', config.get('pubtopic'))
    rabbit.publish(output_data, routing_key=output_key)
elif function_signature == func(data):
    # process like normal
```

## Proposed Solution 2

There is a bit of a debate around how _implicit_ or _explicit_ we should make this behavior. Maybe this is something we specify in the configuration, or maybe it's some runtime magic. Wrt. runtime magic, instead of matching on a specific keyword, or a positional argument, we can perform type inspection to decide if `ergo` is going to provide a context to the application / component, and how it's going to pass it.

For example:
```
def handler.py
---
from ergo.context import ErgoContext
def product(x: int, y: int, ctx: ErgoContext):
    if x % 2 == 0:
       ctx.pubtopic = 'sum'
       return x + y
    return x * y
```

where the application developer can name their variable whatever they want, so long as the type annotation specifies `ErgoContext`. And expanding on this idea, this can be a proper class that can give controlled access to other relevant runtime information.

PR #47, opened 2021-11-30
add context in amqp for dynamic routing

zekenautilus:
## Description

Follows Proposed Solution in https://github.com/nautiluslabsco/ergo/issues/22

If function signature is:
```
def product(x: int, y: int):
    return x * y
```
then functions like normal.

If function signature is:
```
from ergo.context import ErgoContext
def product(ctx: ErgoContext, x: int, y: int):
    if x % 2 == 0:
        ctx.pubtopic = 'sum'
        return x + y
    return x * y
```
then will publish against designated routing key `sum` instead of whatever was set in config.

## Visuals

image

Screen Shot 2021-11-30 at 5 50 43 PM


src/function_invocable.py
```diff
@@ -58,7 +58,7 @@ def func(self, arg: Callable[..., TYPE_RETURN]) -> None:  # type: ignore
         """
         self._func = arg
 
-    def invoke(self, data_in: TYPE_PAYLOAD) -> Generator[TYPE_PAYLOAD, TYPE_PAYLOAD, None]:
+    def invoke(self, data_in: TYPE_PAYLOAD, context) -> Generator[TYPE_PAYLOAD, TYPE_PAYLOAD, None]:
```
zekenautilus, line 61: can default to None so that context is not needed

mattian7741:
Zeke - perhaps you, Ricky and I can jump on a call to discuss this.   Ricky has a specific need for dynamic routing with the work that he is doing on DPI, so he is already working on a solution with my help.   It is similar to what you are proposing here except that it isn't doesn't impose as strong constraints on the function signature.

In effect, the key difference is that context is the first parameter, rather than the last.   And the component configuration yaml provides a signal to ergo that the injected function takes a context.   This signal must be explicit in configuration because we are otherwise forced into an ergo-specific contract.

Perhaps you could join my sync with Ricky tomorrow @ 10:15.   I will add you.

src/function_invocable.py
```diff
@@ -67,6 +67,7 @@ def invoke(self, data_in: TYPE_PAYLOAD) -> Generator[TYPE_PAYLOAD, TYPE_PAYLOAD,
 
         Args:
             data_in (Dict): Contents will be passed to injected function as keyword args.
+            context (Dict): Passes along additional context
```
ahuman-bean, line 70: Should make context a typed object, since its fields are more interesting to document

src/amqp_invoker.py
```diff
@@ -158,5 +160,16 @@ def do_work(self, data_in: TYPE_PAYLOAD) -> Iterable[TYPE_PAYLOAD]:
...
-            yield {'data': data_out, 'key': str(self._invocable.config.pubtopic), 'log': data_in.get('log', [])}
+        data = data_in['data']
+        context = {
+            'subtopic': self._invocable.config.subtopic,
+            'pubtopic': self._invocable.config.pubtopic
+        }
+        for ambiguous_object in self._invocable.invoke(copy.deepcopy(data), copy.deepcopy(context)):
+            if type(ambiguous_object) is tuple:
+                data_out, context_out = ambiguous_object
+                yield {'data': data_out, 'key': context_out['pubtopic'], 'log': data_in.get('log', [])}
```
ahuman-bean, line 171: I'd rather the provided context object be mutable, than to have to return it explicitly.

src/function_invocable.py
```diff
@@ -75,7 +76,11 @@ def invoke(self, data_in: TYPE_PAYLOAD) -> Generator[TYPE_PAYLOAD, TYPE_PAYLOAD,
         if not self._func:
             raise Exception('Cannot execute injected function')
         try:
-            result = self._func(**data_in)
+            num_params = len(inspect.signature(self._func).parameters)
+            if num_params == 2:
+                result = self._func(data_in, context)
+            else:
+                result = self._func(data_in)
```
ahuman-bean, line 83: somewhat hidden within this change is that fact that we're no longing spreading out `data` correct?

ahuman-bean:
> Zeke - perhaps you, Ricky and I can jump on a call to discuss this. Ricky has a specific need for dynamic routing with the work that he is doing on DPI, so he is already working on a solution with my help. It is similar to what you are proposing here except that it isn't doesn't impose as strong constraints on the function signature.
> 
> In effect, the key difference is that context is the first parameter, rather than the last. And the component configuration yaml provides a signal to ergo that the injected function takes a context. This signal must be explicit in configuration because we are otherwise forced into an ergo-specific contract.
> 
> Perhaps you could join my sync with Ricky tomorrow @ 10:15. I will add you.

The placement of `context` in the function signature won't matter if we're leveraging runtime type inspection to provide it anyways

src/amqp_invoker.py
```diff
@@ -158,5 +160,17 @@ def do_work(self, data_in: TYPE_PAYLOAD) -> Iterable[TYPE_PAYLOAD]:
...
+        data = data_in['data']
+        context = {
+            'subtopic': self._invocable.config.subtopic,
+            'pubtopic': self._invocable.config.pubtopic
+        }
+
+        def _helper(g, orig_context):
+            curr_context = copy.deepcopy(orig_context)
+            for _data_out in g(copy.deepcopy(data), curr_context):
+                yield _data_out, curr_context
```
zekenautilus, line 172: can put some of the context logic in function_invocable if that makes more sense? And switch in here on if we get data_out, and context, or just data_out.

zekenautilus:
@ahuman-bean @mattian7741 updated s.t. it matches based on type and filters keywords.

So something like:
```
from src.types import ErgoContext
def zeke_product(dumb_context_name: ErgoContext, x: int, y: int)
    if x % 2 == 0:
        dumb_context_name['pubtopic'] = 'sum'
        return x + y
    return x * y
```

will work as expected (`dumb_context_name`'s position, nor its name matter - we match on the type annotation!) and additionally work if the data payload has additional keys.

For example, publishing:
```
{"data": {"x": 2, "y": 3, "z": 5}}
```
expectedly results in 5 w routing key `sum`!

src/types.py
```diff
@@ -4,7 +4,10 @@
...
 
 """
-from typing import Any
+from typing import Any, Dict
 
 TYPE_RETURN = Any
 TYPE_PAYLOAD = Any
+
+# TODO: make this a proper class!
+ErgoContext = Dict
```
zekenautilus, line 13: not super sure what to put in this class atm. So thinking it's best to alias as Dict for now, until a use case pops up where we might want to promote it from a type alias to its own class.

ahuman-bean: changes to this would be breaking

if we're already committing to allowing `pubtopic`, this could easily just be a dataclass w/ `pubtopic` as a field

using a `Dict` right from the get-go could result in minor versioning annoyances

zekenautilus: It def requires more thought... also i'm not the biggest fan of this currently being export as `src.util.ErgoContext`. it should probably be `ergo.types.ErgoContext`

Also, updating it to be a class that's used across invokers (so not just the amqp invoker) makes it a little weird to include things in the context that are amqp specific (like pubtopic). But at the same time, we already have `pubtopic` at the global level of the config so maybe that's a field that should be consumable by all invokers...

zekenautilus: also another can of worms... should we have invoker specific contexts? Can / should we support passing along any of them to a specific function? ... i.e. instead of `ErgoContext` we pass along `ErgoAmqpContext` which has AMQP specific info (`routing_key` and `exchange`) and maybe a `ErgoHttpContext` which was HTTP specific info (`endpoint` `host)...

zekenautilus: @ahuman-bean updated! still working through my mind, lmk what you think.

ahuman-bean: Then I think it's better to make `Context` an interface and turn `pubtopic` into a concept rather than something concrete. E.g. for HTTP or Console apps, it could be unimplemented and therefore unused.

zekenautilus: I think that's a solid idea. Maybe can be punted on as follow up work? I.e. revisiting what the pubtopic interface is, and so on in the config, and in turn updating the Context to being an interface.

setup.py
```diff
@@ -92,7 +92,8 @@ def run(self):
         'aio-pika',
         'retry',
         'aiomisc',
-        'graphviz'
+        'graphviz',
+        'yarl'
```
zekenautilus, line 96: this `yarl` package is driving me insane... thought it would get picked up by `aio-pika` since it's a dependency for that... but when dev'ing locally as a consumer of ergo i always end up having to install it in my virtualenv

mattian7741 (changes requested):
I am going to look at this in further detail over the weekend and provide a comprehensive response next week.  But this should not be merged as it is.

PR #48, opened 2021-12-09
added context enabler in config and inject context in the invoker

rickyviank:

PR #49, opened 2021-12-12
Context reference and dynamic routing

mattian7741:
This change deprecates #47 and #48.

This is an initial step that uses "context" as it has been discussed.   However, in a subsequent PR - soon to follow - this idea of context will be replaced by a more meaningful implicit binding to any referenced metadata in the communication data protocol for a given namespace.   This PR satisfies the requirement according to the design discussions we have had thus far; and it satisfies both patterns (implicit and explicit argument binding) - because implicit is convenient (defaults) and explicit is necessary.

### DIFF
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -28,7 +28,7 @@ def set_param(host: str, param_key: str, param_val: str) -> str:
 
 def make_error_output(err: Exception) -> Dict[str, str]:
     """Make a more digestable error output."""
-    orig = err.__context__
+    orig = err.__context__ or err
     err_output = {
         'type': type(orig).__name__,
         'message': str(orig),
@@ -104,7 +104,7 @@ class AmqpInvoker(Invoker):
                     data_in.set('error', make_error_output(err))
                     data_in.set('traceback', str(err))
                     data_in.unset('context')
-                    message = aio_pika.Message(body=json.dumps(str(data_in)).encode())
+                    message = aio_pika.Message(body=str(data_in).encode())
                     routing_key = f'{self.queue_name}_error'
                     await self.publish(channel_pool, message, routing_key)
 
@@ -134,7 +134,8 @@ class AmqpInvoker(Invoker):
 
             async for message in queue:
                 async with message.process():
-                    yield Payload(dict(json.loads(message.body.decode('utf-8'))))
+                    data_in = json.loads(message.body.decode('utf-8'))
+                    yield Payload(data_in)
 
     async def publish(self, channel_pool: aio_pika.pool.Pool[aio_pika.RobustChannel], message: aio_pika.Message, routing_key: str) -> None:
         """
@@ -162,5 +163,5 @@ class AmqpInvoker(Invoker):
             payload: Lazily-evaluable wrapper around return values from `self._invocable.invoke`, plus metadata
         """
 
-        for data_out in self._invocable.invoke([data_in.get(self._invocable.config.args[arg]) for arg in self._invocable.config.args]):
+        for data_out in self._invocable.invoke(data_in):
             yield {'data': data_out, 'key': str(data_in.get('context').pubtopic), 'log': data_in.get('log', [])}
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -103,6 +103,7 @@ class AmqpInvoker(Invoker):
                 except Exception as err:  # pylint: disable=broad-except
                     data_in.set('error', make_error_output(err))
                     data_in.set('traceback', str(err))
+                    data_in.unset('context')
                     message = aio_pika.Message(body=json.dumps(str(data_in)).encode())
                     routing_key = f'{self.queue_name}_error'
                     await self.publish(channel_pool, message, routing_key)
--- a/src/amqp_invoker.py
+++ b/src/amqp_invoker.py
@@ -12,7 +12,7 @@ from src.function_invocable import FunctionInvocable
 from src.invoker import Invoker
 from src.types import TYPE_PAYLOAD
 from src.util import extract_from_stack
-
+from src.payload import Payload
 # content_type: application/json
 # {"x":5,"y":7}
 
@@ -93,7 +93,7 @@ class AmqpInvoker(Invoker):
         async with connection, channel_pool:
             async for data_in in self.consume(channel_pool):
                 try:
-                    data_in['context'] = self._invocable.config.copy()
+                    data_in.set('context', self._invocable.config.copy())
 
                     async for data_out in self.do_work(data_in):
                         message = aio_pika.Message(body=json.dumps(data_out).encode())
@@ -101,9 +101,9 @@ class AmqpInvoker(Invoker):
                         await self.publish(channel_pool, message, routing_key=routing_key)
 
                 except Exception as err:  # pylint: disable=broad-except
-                    data_in['error'] = make_error_output(err)
-                    data_in['traceback'] = str(err)
-                    message = aio_pika.Message(body=json.dumps(data_in).encode())
+                    data_in.set('error', make_error_output(err))
+                    data_in.set('traceback', str(err))
+                    message = aio_pika.Message(body=json.dumps(str(data_in)).encode())
                     routing_key = f'{self.queue_name}_error'
                     await self.publish(channel_pool, message, routing_key)
 
@@ -133,7 +133,7 @@ class AmqpInvoker(Invoker):
 
             async for message in queue:
                 async with message.process():
-                    yield dict(json.loads(message.body.decode('utf-8')))
+                    yield Payload(dict(json.loads(message.body.decode('utf-8'))))
 
     async def publish(self, channel_pool: aio_pika.pool.Pool[aio_pika.RobustChannel], message: aio_pika.Message, routing_key: str) -> None:
         """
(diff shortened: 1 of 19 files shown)


src/amqp_invoker.py
```diff
@@ -158,5 +160,6 @@ def do_work(self, data_in: TYPE_PAYLOAD) -> Iterable[TYPE_PAYLOAD]:
         Yields:
             payload: Lazily-evaluable wrapper around return values from `self._invocable.invoke`, plus metadata
         """
-        for data_out in self._invocable.invoke(data_in['data']):
-            yield {'data': data_out, 'key': str(self._invocable.config.pubtopic), 'log': data_in.get('log', [])}
+
+        for data_out in self._invocable.invoke([data_in.get(self._invocable.config.args[arg]) for arg in self._invocable.config.args]):
+            yield {'data': data_out, 'key': str(data_in.get('context').pubtopic), 'log': data_in.get('log', [])}
```
zekenautilus, line 165: I seem to be running into an issue running this locally - the following error centering around `context` (or config) not being json serializable:

image


this is my simple 'passthrough' case where we don't use args and operate on a simple product:

image

zekenautilus: Otherwise these changes work for the other 2 cases:
1. `args` are specified, `context` in function signature
2. `args` are specified, `context` not in function signature

Is this change intended to force us to perform a mapping for all of the arguments in our function? Even if 1:1? Or was that a mistake?

zekenautilus: my guess is that the issue lies in: 
```
[data_in.get(self._invocable.config.args[arg]) for arg in self._invocable.config.args]
```

I think w the 'passthrough' case `self._invocable.config.args` might look something like:
```
{'x': 'x', 'y': 'y'}
```
but data_in might look like:
```
{'data': {
    'x': 2,
    'y': 3,
    }
```
for example, and the `data` nesting throws things off. Also curious if maybe the ordering matters?

src/function_invocable.py
```diff
@@ -119,3 +118,4 @@ def inject(self) -> None:
 
         method_name: str = matches.group(5)
         self._func = getattr(scope, method_name)
+        self._config.args = OrderedDict((signature_arg, self._config.args.get(signature_arg, signature_arg)) for signature_arg in inspect.getfullargspec(self._func)[0])
```
zekenautilus, line 121: changing this from
```
self._config.args.get(signature_arg, signature_arg)
```
to
```
self._config.args.get(signature_arg, f'data.{signature_arg}')
```

seems to make things work! Probably not the change we want, but I hope it illustrates the issue I'm running into.

src/amqp_invoker.py
mattian7741: This is working as expected and context should definitely not be json serializable.

The implicit case (in which the signature has x and y as arguments) assumes that the payload has root-level keys named "x" and "y".   This is intended.   In the situation where we have a root-level "data" key, and we are using data as a function argument, the implicit mapping will also work.   This is consistent with all the components we have built so far.   But in the situation where x and y are sub-keys under a root-level data key, and the function defines x and y in the signature, we are forced to introduced an explicit mapping (ie:  x: data.x, y: data. y) - and this is also intended.

src/function_invocable.py
mattian7741: We definitely don't want to introduce "data" as a prefix in the code like this.   But the issues you are running into are expected and intended.   This is how it is supposed to work.

src/amqp_invoker.py
mattian7741: Of course - more elegant error handling is definitely in order.

zekenautilus: hmm if context isn't json serializable and we're setting it in `data` we won't be able to ever write to the error_queue when an exception occurs - that seems like a significant regression

mattian7741: Oh - I see what you mean now @zekenautilus.   There is an error with data_out because context is included (which it shouldn't be) and that is why the json serializable issue is coming up.   We definitely do not want to serialize config/context - but we do want to omit config/context from the outbound message.  I will correct that.

mattian7741:
@zekenautilus @zachschubert - I fixed the primary cause of the json serialization issue by removing config/context from the outbound error message.   This is an appropriate fix under the circumstances but additional changes in the near future will eliminate this construct entirely.   For now the abstraction is properly maintained.

src/amqp_invoker.py
```diff
@@ -93,15 +93,18 @@ async def get_channel() -> aio_pika.Channel:
...
                         await self.publish(channel_pool, message, routing_key=routing_key)
 
                 except Exception as err:  # pylint: disable=broad-except
-                    data_in['error'] = make_error_output(err)
-                    data_in['traceback'] = str(err)
-                    message = aio_pika.Message(body=json.dumps(data_in).encode())
+                    data_in.set('error', make_error_output(err))
+                    data_in.set('traceback', str(err))
+                    data_in.unset('context')
+                    message = aio_pika.Message(body=json.dumps(str(data_in)).encode())
```
zachschubert, line 107: `json.dumps(str(data_in))` doubly serializes data_in, so that you have to call `json.loads` twice on that string to get a dict back. I added a new test in https://github.com/nautiluslabsco/ergo/pull/51 to cover this path.

zekenautilus: nice catch! so this should just be `body=str(data_in).encode()` ?

zekenautilus:
> @zekenautilus @zachschubert - I fixed the primary cause of the json serialization issue by removing config/context from the outbound error message. This is an appropriate fix under the circumstances but additional changes in the near future will eliminate this construct entirely. For now the abstraction is properly maintained.

I think that may be fine, except for the cases where components modify the input topic and respond w it, or a copy of it. Since they're being passed `context` in `data` then their outbound message will also have `context` - and when we try to publish that message we're still going to see the json serialization error. Curious what's the argument for not allowing the `context` to be json serializable?

referencing: https://github.com/nautiluslabsco/qed/blob/master/components/parser/handler_parser.py#L13-L24

alternatively we could just leave those components at the version of ergo they're currently deployed with...

src/amqp_invoker.py
zachschubert: that's right

zachschubert (commented):
@mattian7741 Was it your intention to merge this PR before implementing the other changes you described [here](https://github.com/nautiluslabsco/ergo/pull/51#issuecomment-994234343)? I believe they need to be released together. Otherwise, these changes would break any component which sits downstream from another component.

src/amqp_invoker.py
mattian7741: The "data" key is an artifact of a specific communication protocol - on a different ergo system payloads might use a completely different structure.   But the idea mapping root level keys to arguments can work implicitly.   Therefore, if the signature of the function includes "data", and data is at the root of the data-structure, the argument can be bound.   Implicit mapping of keys that are deeper in the data-structure, is problematic and the path to these deeper keys should be provided explicitly.   An exception to this can be made in the namespace - for example, we could specify that for the namespace "nautilus", in which a specific communication protocol is used consistently, that "data" represents root, and in that case keys that live under data will map implicitly to function arguments.

mattian7741: Regarding the error queue behavior - that needs to be revisited anyway, but I don't believe this PR interferes in any way with current behavior EXCEPT if we previously implicitly mapped subkeys of "data" to function arguments.

mattian7741:
> > @zekenautilus @zachschubert - I fixed the primary cause of the json serialization issue by removing config/context from the outbound error message. This is an appropriate fix under the circumstances but additional changes in the near future will eliminate this construct entirely. For now the abstraction is properly maintained.
> 
> @mattian7741 I think that may be fine, except for the cases where components modify the input topic and respond w it, or a copy of it. Since they're being passed `context` in `data` then their outbound message will also have `context` - and when we try to publish that message we're still going to see the json serialization error. Curious what's the argument for not allowing the `context` to be json serializable?
> 
> referencing: https://github.com/nautiluslabsco/qed/blob/master/components/parser/handler_parser.py#L13-L24
> 
> alternatively we could just leave those components at the version of ergo they're currently deployed with...
> 
> also some possibly relevant python libs:
> 
> * https://docs.python.org/3/library/dataclasses.html
> * https://lidatong.github.io/dataclasses-json/

The first reason for not making context json serializable is because it isn't necessary - context was never included in any outbound messages (and shouldn't have been), and it isn't included now either.   The second reason is that context (or the idea of a context-like object for meta programming against ergo from within a component) is a reference to collection of runtime object instances that can be manipulated directly - serializing and deserializing these would be unnecessary and unproductive.   So that still leaves the question: why don't we want to include context in the outbound message?   And again, the idea of doing so never even came up until we accidentally attempted it and got an error.   There is no reason to include it in any outbound messages.   There are certain fields that belong to that context that might be/are relevant in the outbound message, but they are sourced explicitly (pubtopic, for example).   The only time we would ever want to serialize context, is if we were running a component written in another language (eg. node.js) using the python ergo runtime, or perhaps if we wanted to log context, or for some other reason not related to the sending and receiving of messages.

mattian7741:
> @mattian7741 Was it your intention to merge this PR before implementing the other changes you described [here](https://github.com/nautiluslabsco/ergo/pull/51#issuecomment-994234343)? I believe they need to be released together. Otherwise, these changes would break any component which sits downstream from another component.

My PR was intended to leave existing behavior unchanged while also introducing an option to include reference context in the function signature.   The only exception is if we were previously implicitly mapping subkeys of "data" to signature arguments - and we necessarily have to stop doing that.   But the expectation (for now) is that components running on amqp networks will receive messages with their data in a "data" key, and publish messages with data housed under the "data" key.    This hard constraint will be removed soon (in a subsequent PR) - but for now it satisfies the (artificial) communication standard that data is carried under a data key.   The asymmetry here is that while we conventionally use data as the transport envelope and on the "publish" side we are hard coding this "data" key envelope, on the "receive" side, we are now also able to handle more general payloads that don't use this "data" construct.  But since we are still using the "data" construct in our messages, we don't experience this asymmetry in  our specific usecase.

zekenautilus (commented):
cool merging this once the double serialization issue is fixed and the changes in 
https://github.com/nautiluslabsco/ergo/pull/51 are merged in so that tests pass again!

zekenautilus (approved):
👍

PR #51, opened 2021-12-14
fix tests

zachschubert:
This PR applies the changes in https://github.com/nautiluslabsco/ergo/pull/49 to the http path, fixes some minor bugs in the error handling path, and tweaks the assertions in the amqp tests.


src/amqp_invoker.py
```diff
@@ -28,7 +29,7 @@ def set_param(host: str, param_key: str, param_val: str) -> str:
 
 def make_error_output(err: Exception) -> Dict[str, str]:
     """Make a more digestable error output."""
-    orig = err.__context__
+    orig = err.__context__ or err
```
zachschubert, line 32: `err.__context__` can be None

src/serialization.py
```diff
@@ -0,0 +1,21 @@
+import json
+from typing import Any
+
+from src.config import Config
+from src.payload import Payload
+from src.topic import Topic
+
+
+def serialize(data: Any) -> str:
```
mattian7741, line 9: I am not opposed to breaking serialization out as an encapsulated function, but doing so hides underlying issues.   The first is that we don't want to serialize config.   We might want to for some reason at some point, but not in this case.   And there is no reason why topic and payload shouldn't be serialized with __str__ in general.   So the introduction of this serialization class in some ways condones errors further up stream.

mattian7741:
One other point - I know that the outbound (returned) data is still being mapped into a data element and I will fix that in a subsequent PR.   I didn't want to include too much in the first PR - rather do it in bite sized chunks and keep them reviewable.   Also wanted to keep the scope of the change to the problem we were already discussing.   But the outbound data binding and a host of other changes are on the list for refactor very soon.

mattian7741:
The primary upstream issue here is that where data_in is serialized in the exception block, it should not include config/context.   That has to be omitted from the outbound payload.

src/serialization.py
zachschubert: ergo is trying to serialize Config [here](https://github.com/nautiluslabsco/ergo/blob/3ea4f07c4251f7a5fb28f3797dd1290ff2dbeca1/src/amqp_invoker.py#L107) to publish to the error queue. Shall I pop it from data_in instead?

zachschubert: Oh never mind just saw https://github.com/nautiluslabsco/ergo/pull/51#issuecomment-994263410

zachschubert:
@mattian7741 I merged your last commit, and removed the deprecated `serializer.serialize`.

src/function_invocable.py
```diff
@@ -76,7 +76,8 @@ def invoke(self, data_in: TYPE_PAYLOAD) -> Generator[TYPE_PAYLOAD, TYPE_PAYLOAD,
         if not self._func:
             raise Exception('Cannot execute injected function')
         try:
-            result = self._func(*data_in)
+            args = [data_in.get(self.config.args[arg]) for arg in self.config.args]
```
zekenautilus, line 79: ah so this is moved into function_invocable so that the paradigm works w other invokers? That makes a bit more sense to me!

zekenautilus (approved):
👍

PR #53, opened 2022-01-03
add docs, mv module, lint tests

zekenautilus:
## Description

- added docs! currently built locally, but signed up for `readthedocs` and should be able to host there!
- some initial linting ci tests to verify code style is enforced pre merge
- renamed module from `src` to `ergo`

## Visuals

Uploading Screen Recording 2022-01-03 at 9.05.00 AM.mov…


.circleci/config.yml
```diff
@@ -26,8 +39,24 @@ jobs:
             . venv/bin/activate
             make test
             codecov
+      - run:
+          name: "Enforce PEP8"
+          command: |
+            . venv/bin/activate
+            autopep8 --aggressive --aggressive --recursive ./ergo --diff --exit-code
```
ahuman-bean, line 46: aggressive twice?

ahuman-bean (approved):
⭐

.circleci/config.yml
anwar-nautilus: just means be extra aggressive

zekenautilus: i legit thought it was a typo, but this is the intended usage:

image

PR #57, opened 2022-01-13
Schubert/transactions pd 613

zachschubert:
1. Add a (Transaction) Stack class with push and pop methods.
2. Add a Context class with methods for opening and closing transactions.
3. Refactor the Payload class as a datadict, adding a new `jsons` dependency for serializing and deserializing. The primary benefit of this is change is that jsons can use type info to decode strings into Payload objects. With this we can e.g. add a new attribute to Stack (with a default value), and payload.decodes & payload.encodes will incorporate it automatically. Then later we can delete the attribute and still safely decode old messages because jsons will automatically filter what it passes to the Stack constructor.
4. Add a bunch of test util improvements to support the new transactions tests in `test_stack.py`.


ergo/function_invocable.py
```diff
@@ -118,5 +124,17 @@ def inject(self) -> None:
...
+                func = func.__call__
+            self._func = func
+
+            params = {}
+            for name, info in inspect.signature(self._func).parameters.items():
+                default = info.default
+                if default is inspect.Parameter.empty:
+                    default = None
+                params[name] = default
+            self._config.parameters = params
```
zachschubert, line 140: this ended up being unrelated, but it allows ergo to inject class/instance methods and functions, and stores the correct default kwarg values.

ergo/config.py
```diff
@@ -26,28 +26,28 @@ def __init__(self, config: Dict[str, str]):
         self._exchange: Optional[str] = config.get('exchange')
         self._protocol: str = config.get('protocol', 'stack')  # http, amqp, stdio, stack
         self._heartbeat: Optional[str] = config.get('heartbeat')
-        self._args: OrderedDict[str, str] = config.get('args', {})
+        self._parameters: Optional[OrderedDict] = None
```
zachschubert, line 29: see https://github.com/nautiluslabsco/ergo/pull/57/files#r786050388

ergo/function_invocable.py
```diff
@@ -118,5 +124,18 @@ def inject(self) -> None:
...
+                func = func.__call__
+            self._func = func
+
+            params = {}
+            for name, info in inspect.signature(self._func).parameters.items():
+                default = info.default
+                if default is inspect.Parameter.empty:
+                    default = None
+                params[name] = default
+            self._config.args = params
```
zachschubert, line 141: this ended up being unrelated, but it allows ergo to inject class/instance methods and functions, and stores the correct default kwarg values.

ergo/config.py
```diff
@@ -20,13 +20,13 @@ def __init__(self, config: Dict[str, str]):
         """
         self._func: str = config['func']
         self._namespace: Optional[str] = config.get('namespace', 'local')
-        self._pubtopic: Topic = PubTopic(config.get('pubtopic'))
-        self._subtopic: Topic = SubTopic(config.get('subtopic'))
+        self._pubtopic: str = config.get('pubtopic')
+        self._subtopic: str = config.get('subtopic')
```
zachschubert, line 24: we need to store the raw pubtopic here, because that's what we want to initialize Context objects with. it will also be convenient later to have a [method](https://github.com/nautiluslabsco/ergo/pull/58/files#diff-cf19c12d02789b27dafc4ccf55362134954ef4710a5d6b70ed6450bb946dc277R141-R143) for creating dynamic subscriptions that automatically reformats routing_keys by converting them to `PubTopic`s

ergo/function_invocable.py
```diff
@@ -76,12 +77,18 @@ def invoke(self, data_in: TYPE_PAYLOAD) -> Generator[TYPE_PAYLOAD, TYPE_PAYLOAD,
...
+            for param, default in self._config.args.items():
+                if param == "context":
+                    kwargs["context"] = ctx
+                else:
+                    kwargs[param] = data_in.get(param, default)
+            results = self._func(**kwargs)
+            if not inspect.isgenerator(results):
+                results = [results]
+            for result in results:
+                yield Payload(data=result, stack=ctx._stack, key=ctx.pubtopic)
```
zachschubert, line 91: `ctx._stack` and `ctx.pubtopic` are both mutable, and we expect them to be updated from one result to the next. thus IMO it's safer to keep the context object local to this method, and to only yield data that we can safely copy out of it.

ergo/payload.py
```diff
@@ -1,68 +1,45 @@
...
-
-    def unset(self, key: str) -> str:
-        """Summary.
-
-        Args:
-            key (str): Description
-            value (str): Description
+@dataclass
+class Payload:
+    data: Any = field(default=None)
```
mattian7741, line 15: Should context not be part of this list?

ergo/payload.py
```diff
@@ -1,68 +1,45 @@
...
+    key: Optional[str] = None
```
mattian7741, line 16: What does key represent?

ergo/payload.py
```diff
@@ -1,68 +1,45 @@
...
-        self._data[key] = value
-
-    def unset(self, key: str) -> str:
-        """Summary.
-
-        Args:
-            key (str): Description
-            value (str): Description
+@dataclass
+class Payload:
```
mattian7741, line 14: Payload by its name represents the data element - the introduction of the other fields here changes its meaning.   That is fine, and I'm not terribly worried about the name.   I am a bit concerned with the ambiguity that is developing around payload, metadata, context, data_in, data_out, etc.   All of these concepts should be converging into a common class and it feels as if we're mapping back and forth between these concepts.

ergo/config.py
mattian7741: I'm not following the logic here (I'm sure it's my fault).   This one might be easier to discuss in person.

mattian7741 (commented):
Left a few comments here.   Worth some additional discussion before finalizing.

ergo/stack.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Optional
+
+from ergo.util import uniqueid
+
+
+@dataclass
+class Stack:
```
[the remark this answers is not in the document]
mattian7741: How/where is the current stack included in the message data?

ergo/function_invocable.py
mattian7741: I agree with this.   I don't think the Context constructor should take pubtopic and stack.   Pubtopic is just one of many potential input parameters to creating a context - Using a context factory that can source information from anywhere to create a valid context makes more sense to me, or accepting a dictionary.   Or it (or a factory on its behalf) could accept immutable config and inbound message to construct mutable context.

mattian7741 (approved):

PR #58, opened 2022-01-21
Schubert/instance subscriptions pd 611 pd 612

zachschubert:
1. Assign component instances unique IDs provided by `utils.uniqueid`.
2. Bind the AMQPInvoker's main consumer queue to that instance ID.
3. Add a `Stack.data` attribute to store unstructured (at least for now) key-value data.
4. Use the above to implement `Context.request` and `Context.respond`. `request` takes a topic and optional data, and generates an outbound payload to be published with that topic in a temporary transaction. `respond` takes arbitrary data, and generates an outbound payload to be published with the requester's instance ID as the routing key. The responder's normal pubtopic isn't included in that routing key, nor is the transaction ID associated with request. This way, the responder doesn't have to be responsible for knowing which other components might have received the same request, or what side effects might follow if they received each other's responses.

### DIFF
--- a/ergo/amqp_invoker.py
+++ b/ergo/amqp_invoker.py
@@ -95,6 +95,8 @@ class AmqpInvoker(Invoker):
             async for data_in in self.consume(channel_pool):
                 try:
                     async for data_out in self.do_work(data_in):
+                        if data_in.stack and data_in.stack.get_callback_key():
+                            data_out.key = f"{data_out.key}.response"
                         message = aio_pika.Message(body=encodes(data_out).encode('utf-8'))
                         routing_key = str(PubTopic(data_out.key))
                         await self.publish(channel_pool, message, routing_key=routing_key)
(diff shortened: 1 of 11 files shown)


ergo/amqp_invoker.py
```diff
@@ -127,13 +129,19 @@ async def consume(self, channel_pool: aio_pika.pool.Pool[aio_pika.RobustChannel]
             queue = await channel.declare_queue(name=self.queue_name)
             queue_error = await channel.declare_queue(name=f'{self.queue_name}_error')
 
-            await queue.bind(exchange=exchange, routing_key=str(SubTopic(self._invocable.config.subtopic)))
+            await self.subscribe(self._invocable.config.subtopic, queue, exchange)
+            response_topic = f"{self.instance_id}.response"
```
mattian7741, line 133: no ".response" because an instance should receive messages that are destined for it whether they are a response or not.

Also - the instance_id is not enough on its own because it could collide with other id-like keys in the topic that refer to other things (not instances).  So a type prefix has to be added eg.  f"instanceid_{self.instance_id}"

Incidentally - an upcoming future ticket will introduce topic compression.  Rabbitmq limits topics to 255 characters so as topics become more elaborate, we will start approaching that limit.   A just-in-time compression step has to be injected on the output and input to compress and decompress topics accordingly.   In the past I did this with a memory-resident map of base36 numbers mapped to specific topic key tokens.   Example: 0, 1, 2, ..., 9, a, b, c, ..., x, y, z, 11, 12, etc.

ergo/amqp_invoker.py
```diff
@@ -95,6 +95,8 @@ async def get_channel() -> aio_pika.Channel:
             async for data_in in self.consume(channel_pool):
                 try:
                     async for data_out in self.do_work(data_in):
+                        if data_in.stack and data_in.stack.get_reply_to():
+                            data_out.key = f"{data_out.key}.response"
```
mattian7741, line 99:
```
f"{data_in.stack.get_reply_to()}.{data_out.key}.response"
```

ergo/context.py
```diff
@@ -1,12 +1,67 @@
...
+            self._stack.set_reply_to(instance_id())
+            self._results_stream.send(Payload(data=kwargs or None, key=f"{topic}.request", stack=self._stack))
+
+    def respond(self, data=None, **kwargs):
+        """
+        See docstring for Context.request.
+
+        """
+        if self._stack.get_reply_to() == instance_id():
+            # We must be inside a recursive request stack, and the current transaction is associated with a request call
```
mattian7741, line 47: Let's schedule time for this one.   I will set something up

ergo/amqp_invoker.py
```diff
@@ -95,6 +95,8 @@ async def get_channel() -> aio_pika.Channel:
             async for data_in in self.consume(channel_pool):
                 try:
                     async for data_out in self.do_work(data_in):
+                        if data_in.stack and data_in.stack.get_reply_to():
```
mattian7741, line 98: The condition here should be assessed at a lower level. 
```
if inbound_topic contains "request":
   append "response" and instance_id to outbound topic
```

PR #60, opened 2022-02-02
Schubert/reorganize amqp invoker pd 612

zachschubert:


ergo/amqp_invoker.py
```diff
@@ -70,82 +73,86 @@ def start(self) -> int:
...
         """
         Re-acquires handles to `channel`, `exchange`, and `queue` before publishing message.
 
         Parameters:
-            channel_pool: Pool of valid `Channel` handles
+            channel: aio_pika.Channel
             message: Message to be published
             routing_key: Exchange-level routing discriminator
         """
-        async with channel_pool.acquire() as channel:
```
ahuman-bean, line 146: This was done intentionally to reacquire the handle before publishing, especially if a very long time has passed before `publish` actually gets called

ergo/amqp_invoker.py
```diff
@@ -70,82 +73,86 @@ def start(self) -> int:
...
+            aio_pika.pool.Pool(get_channel, max_size=CHANNEL_POOL_SIZE, loop=loop)
         async with channel_pool.acquire() as channel:
-            await channel.declare_exchange(name=self.exchange_name, type=aio_pika.ExchangeType.TOPIC, passive=False, durable=True, auto_delete=False, internal=False, arguments=None)
+            exchange = await channel.declare_exchange(
+                name=self.exchange_name, type=aio_pika.ExchangeType.TOPIC, passive=False, durable=True,
+                auto_delete=False, internal=False, arguments=None)
+            error_queue = await channel.declare_queue(name=self.error_queue_name)
+            await error_queue.bind(exchange=exchange, routing_key=self.error_queue_name)
+            component_queue = await channel.declare_queue(name=self.component_queue_name)
+            await component_queue.bind(exchange=exchange, routing_key=str(SubTopic(self._invocable.config.subtopic)))
```
ahuman-bean, line 104: As per https://github.com/nautiluslabsco/ergo/pull/41 (see PR description), for both `queue.bind` and `channel.acquire`, I found it necessary to have to refresh the handle for long-running ergo components. This was the main reasoning behind it being so close to the logic inside `consume`, as opposed to being called once in `run`. If you are able to test locally with a component that takes greater than say, 30 minutes to complete, without running into any exceptions, then perhaps it was a transient error that needs to be reproduced. From my personal experience however, since those changes, the long-running component I was working on has been consistently stable both deployed and locally (which was not the case before).

ahuman-bean: But generally pulling out the queue binding into a helper func makes sense, but it will probably still need to be called inside of `consume`

zachschubert: reacquiring a channel to publish makes sense. but `consume` in master only declares queues once before starting the infinite loop, so it seems to me that moving those calls into `run` would be equivalent. in any case, i just started a long-running component test, so i'll get back to you with the results in a bit

ergo/amqp_invoker.py
zachschubert: ⚡

ahuman-bean (approved):
👍

ergo/amqp_invoker.py
zekenautilus: were the tests w long-running components successful?

ergo/amqp_invoker.py
```diff
@@ -72,80 +75,70 @@ def start(self) -> int:
...
-
-    async def publish(self, channel_pool: aio_pika.pool.Pool[aio_pika.RobustChannel], message: aio_pika.Message, routing_key: str) -> None:
+            async for amqp_message in queue:
+                amqp_message = cast(aio_pika.IncomingMessage, amqp_message)
+                ergo_message = decodes(amqp_message.body.decode('utf-8'))
+                if self._terminating:
+                    break
+                with self.defer_termination():
+                    amqp_message.ack()
+                    await self.handle_message(ergo_message, channel)
```
zekenautilus, line 116: this changes our behavior to acking prior to processing the message, correct? I think this direction makes sense, so long as we're fine with losing messages due to system errors. For example - if we ack a message, start processing (maybe pulling in weather data for example, something memory intensive) and get OOMkilled we're not going to requeue the message, or send it to the error queue.

I think acking early is fine however (after decoding the data and such).

zekenautilus: ahhh - so the proposed behavior here is that defer_termination() catches the signal? I still think that it's possible w existing components that a system level issue will occur that cannot be caught (such as being OOMkilled) but handling that is probably the responsibility of the infrastructure that's running the components, not the ergo library.

ergo/invoker.py
```diff
@@ -31,3 +35,20 @@ def start(self) -> int:
...
         yield from self._invocable.invoke(message_in)
+
+    @contextmanager
+    def defer_termination(self):
+        """
+        Use this context manager to temporarily postpone SIGTERM handling.
+        """
+        prev_handler = signal.signal(signal.SIGTERM, self._sigterm_handler)
+        try:
+            yield
```
zekenautilus, line 46: does this pass contextmanager pass the message into the error_queue? can you elaborate on this? (not that the code is unclear, just that I'm unfamiliar w python contextmanagers + generators)

ergo/amqp_invoker.py
ahuman-bean: @zekenautilus We discussed offline, so we should probably update the description or sync up so the intent is clear. Basically, we don't want to drop messages where the intent is `SIGTERM`, as opposed to `SIGKILL`. For the latter, it indicates an error that probably needs some intervention, and we wouldn't want to re-queue anyways.

ergo/invoker.py
zachschubert: if ergo receives a sigterm while we're inside this [block](https://github.com/nautiluslabsco/ergo/blob/a5fe62f1a95885995ba26455a13e57049d59d328/ergo/amqp_invoker.py#L117-L119), it'll finish the block before it shuts down. so there wouldn't be any error to publish, at least on account of the sigterm itself. if we get the sigterm [here](https://github.com/nautiluslabsco/ergo/blob/a5fe62f1a95885995ba26455a13e57049d59d328/ergo/amqp_invoker.py#L116), then ergo won't ack the message, and the broker will send it to another instance

ergo/amqp_invoker.py
zachschubert: they were successful. i also intend to run some new ones before merging, just in case

mattian7741 (approved):

PR #61, opened 2022-02-03
add instance queue

zachschubert:


test/integration/amqp/test_scope.py
```diff
@@ -18,8 +18,8 @@ def test_simple_scope(component):
     component.send()
     scopes = [component.consume()["scope"] for _ in range(2)]
     initial_scope, new_scope = sorted(scopes, key=stack_depth)
-    assert initial_scope is None
-    assert new_scope["parent"] is None
+    assert initial_scope["parent"] is None
+    assert new_scope["parent"] == initial_scope
```
zachschubert, line 22: all of these test changes are accounting for the new non-null root scope

ergo/function_invocable.py
```diff
@@ -77,18 +80,47 @@ def invoke(self, data_in: Message) -> Generator[Message, None, None]:
...
+                if isinstance(data_out, Envelope):
+                    envelope = data_out
+                    data_out = envelope.data
+                scope = ctx._scope
+                if Topic(f"{self.config.subtopic}.{instance_id()}").overlap(Topic(scope.reply_to)):
+                    # The current scope was initiated in conjunction with a request that was addressed to this
+                    # component or instance. We assume that by handling this message we've resolved
+                    # the request, and may exit the current scope before proceeding. This frees handlers from
+                    # the need to manually exit scope after receiving a request, or else publishing a message
+                    # which will be routed back to them unto eternity.
```
zachschubert, line 107: i know that reply_to doesn't imply request, but this was the hardest comment i've ever had to write, and i couldn't think of a way to do it except in those terms.

ergo/function_invocable.py
```diff
@@ -77,18 +80,47 @@ def invoke(self, data_in: Message) -> Generator[Message, None, None]:
...
                 else:
-                    kwargs[param] = data_in.get(param, default)
+                    kwargs[param] = message_in.get(param, default)
             results = self._func(**kwargs)
             if not inspect.isgenerator(results):
                 results = [results]
-            for result in results:
-                yield Message(data=result, scope=ctx._scope, key=ctx.pubtopic)
+            for data_out in results:
+                # As a rule, we don't publish a null payload unless it was returned in an Envelope.
```
zachschubert, line 94: this is a rule that i just made up, but it seems good, right? this way a handler can decline to publish by returning nothing, but if it wants to publish a null payload it can do so by explicitly returning `context.envelope(None)`

mattian7741 (approved):

PR #62, opened 2022-02-04
[PD-614] feat(ergo): in message scope

jaechoe:
Expose methods via Context to be able to set/get scope.data.


ergo/context.py
```diff
@@ -14,3 +14,11 @@ def initiate_scope(self):
...
         if self._scope:
             self._scope = self._scope.parent
+
+    def get_scope_data(self) -> dict:
+        if self._scope:
+            return self._scope.data
+
+    def set_scope_data(self, new_data: dict):
+        if self._scope:
+            self._scope.data = new_data
```
mattian7741, line 24: These are fine, but I think they're logically private.   The public methods we need most are more like:
```
context.scope.set(<deep_key>, <value>) => None
context.scope.get(<deep_key>) => <value>
```
@zachschubert - just have to figure out - in the contract - how to disambiguate in-message scope data from shared storage scope data

jaechoe: Wouldn't that be depends on whether we want `scope` to be public vs private from the perspective of the components? If we want the components to have direct access to `scope`, `context.scope` vs `context.getScope()`, I think your suggestion make sense to me (except I'm not sure what `<deep_key>` is referring to). Also, do we want the components to get and set each value or do we want to offload the management of "data" object to each components?

jaechoe:
@zachschubert this should be ready to be merged into your PR (https://github.com/nautiluslabsco/ergo/pull/57)

PR #67, opened 2022-02-28
remove illegal slashes from amqp queue names

zachschubert:


jaechoe (approved):

zekenautilus (approved):
👍  cc @ahuman-bean since this might break autoscaling metrics.

also maybe version should be incremented?

ahuman-bean:
> 👍 cc @ahuman-bean since this might break autoscaling metrics.
> 
> also maybe version should be incremented?

Should be okay to merge (as far as autoscaling metrics are concerned, we've not decided on whether or not queue names will be configured explicitly or enumerated automatically)

PR #68, opened 2022-03-01
declare and consume amqp queues with dedicated channels

zachschubert:
This is a hail mary attempt to solve that DuplicateConsumerTag error that's been showing up in the [telenor connector](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fecs$252Fqed-telenor_api_connector-qa/log-events/ecs$252Fqed-telenor_api_connector-qa$252F44de56c219ef498ca557a1eb9e9cffe5). Digging through the [code](https://github.com/mosquito/aiormq/blob/master/aiormq/channel.py#L478), I haven't figured out specifically how we could be triggering that exception. But I did notice that we're declaring both queues with the same channel, and that we'll end up using that channel in both consumer loops. Channels aren't supposed to be shared btw threads, so even though I can't figure out how it could be responsible for this bug, I still think we need to change it.

### DIFF
--- a/ergo/amqp_invoker.py
+++ b/ergo/amqp_invoker.py
@@ -93,17 +93,21 @@ class AmqpInvoker(Invoker):
         async def get_channel() -> aio_pika.Channel:
             return await connection.channel()
 
-        # Pool for consuming and publishing
-        channel_pool: aio_pika.pool.Pool[aio_pika.RobustChannel] = aio_pika.pool.Pool(get_channel, max_size=CHANNEL_POOL_SIZE, loop=loop)
-        async with channel_pool.acquire() as channel:
-            await channel.declare_exchange(name=self.exchange_name, type=aio_pika.ExchangeType.TOPIC, passive=False, durable=True, auto_delete=False, internal=False, arguments=None)
-            await channel.declare_queue(name=self.error_queue_name)
-            await self.bind_queue(self.error_queue_name, self.error_queue_name, channel)
-            component_queue = await channel.declare_queue(name=self.component_queue_name)
-            await self.bind_queue(self.component_queue_name, self._invocable.config.subtopic, channel)
-            instance_queue = await channel.declare_queue(name=self.instance_queue_name, exclusive=True)
-            await self.bind_queue(self.instance_queue_name, instance_id(), channel)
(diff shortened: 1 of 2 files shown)


ergo/amqp_invoker.py
```diff
@@ -93,17 +93,22 @@ async def run(self, loop: asyncio.AbstractEventLoop) -> aio_pika.RobustConnectio
...
+        await component_channel.declare_exchange(name=self.exchange_name, type=aio_pika.ExchangeType.TOPIC, passive=False,
+                                                 durable=True, auto_delete=False, internal=False, arguments=None)
+        await component_channel.declare_queue(name=self.error_queue_name)
+        await self.bind_queue(self.error_queue_name, self.error_queue_name, component_channel)
+        component_queue = await component_channel.declare_queue(name=self.component_queue_name)
+        component_queue = cast(aio_pika.Queue, component_queue)
+        await self.bind_queue(self.component_queue_name, self._invocable.config.subtopic, component_channel)
+
+        instance_channel = await get_channel()
+        instance_queue = await instance_channel.declare_queue(name=self.instance_queue_name, exclusive=True)
```
ahuman-bean, line 106: can be simplified `instance_queue: aio_pika.Queue` instead of `cast`

ahuman-bean: same w/ `component_queue`

ahuman-bean:
Have you checked the behavior when we begin consuming? Afaik, that's when the consumer tag actually gets bound

ergo/amqp_invoker.py
zachschubert: for some reason RobustChannel.declare_queue returns AbstractRobustQueue, so that `instance_queue: aio_pika.Queue` gets flagged by mypy.

zachschubert:
> Have you checked the behavior when we begin consuming? Afaik, that's when the consumer tag actually gets bound

I have, and it is. It's possible to set that tag manually, but I haven't managed to produce a scenario where we do. Barring that, the only way to get duplicate tags would be to get duplicate `UUID(int=getrandbits(128), version=4)`, which is just supposed to be impossible.

PR #69, opened 2022-03-07
correctly apply configuration args mappings

zachschubert:
I just noticed that ergo 0.6.0a dropped support for custom `args` mapping in configuration yamls. This PR fixes that, and adds an integration test to cover it.


zekenautilus (approved):
⚡ nice catch and nice on adding tests!

PR #70, opened 2022-03-10
refactor AmqpInvoker with kombu

zachschubert:
The intermittent `DuplicateConsumerTag` errors affecting some QED API connectors appears to me to be another bug in `aio-pika`'s self-healing connection code. I haven't thought of any other way around it except to substitute `aio-pika` entirely with an alternative. `kombu` has comparable (hopefully aside from the bugginess) connection restoration utilities, so I'm giving it a shot here. If it works well, I'll follow up with another PR to refactor `kombu` into the http gateway and test kit.

This implementation doesn't completely solve our long-running tasks issue, but it seems to get us a lot closer. The problem is that while I can restore closed connections and channels in between consumer loop iterations, I haven't managed to do it at message ack time. So an hour-long task will finish, but the message will end up getting re-enqueued because we'll try to ack it on a closed connection/channel. On the positive side, the connection is being closed in this case because we're exceeding a broker-side 15 minute ack timeout. So presumably we can finish solving this with a rabbitmq.conf change.

NOTE I had to rename `test/integration/amqp` because `kombu` adds a conflicting `amqp` dependency.


ergo/amqp_invoker.py
```diff
@@ -47,123 +50,87 @@ class AmqpInvoker(Invoker):
...
+    def _producer(self) -> kombu.Producer:
+        with producers[self._connection].acquire(block=True) as conn:
+            yield conn
+
+    def _sigterm_handler(self, *args):
+        self._terminating.set()
+        self._pending_invocations.acquire(blocking=True, timeout=TERMINATION_GRACE_PERIOD)
+        self._connection.close()
+        signal.signal(signal.SIGTERM, 0)
+        signal.raise_signal(signal.SIGTERM)
```
ahuman-bean, line 136: is re-raising `SIGTERM` necessary? I thought we pass control back to the OS when we leave the handler

setup.py
```diff
@@ -96,6 +96,7 @@ def run(self):
         'pydash',
         'jsons',
         'quart',
+        'kombu',
```
ahuman-bean, line 99: remove aio-pika as well?

ergo/amqp_invoker.py
```diff
@@ -47,123 +50,87 @@ class AmqpInvoker(Invoker):
...
+            try:
+                for message_out in self.invoke_handler(message_in):
+                    routing_key = str(PubTopic(message_out.key))
+                    self._publish(message_out, routing_key)
+            except Exception as err:  # pylint: disable=broad-except
+                message_in.error = make_error_output(err)
+                message_in.traceback = str(err)
+                self._publish(message_in, self._error_queue.name)
+            finally:
+                # TODO figure out why this sometimes raises amqp.exceptions.RecoverableConnectionError on SIGTERM
```
ahuman-bean, line 110: could you elaborate a bit more on this?

setup.py
zachschubert: we still need it for the http gateway and test kit, for now

ergo/amqp_invoker.py
zachschubert: that's correct. i added the re-raise for convenience, because otherwise the consumer will re-declare all the queues after my [test teardown code](https://github.com/nautiluslabsco/ergo/blob/3b349c86c6c0bb8664f1bdce818ef9703985e354/test/integration/utils/amqp.py#L104-L105) deletes them. obv that convenience might not be justified. do you think i should remove that?

zachschubert: i say convenience because i like having fewer variables in the amqp console when i'm debugging failed tests

ahuman-bean: fair enough

ergo/amqp_invoker.py
zachschubert: never mind. turns out i just forgot that ctrl-c sends SIGINT and not SIGTERM

ahuman-bean (approved):
Nice! Much cleaner interface

ergo/amqp_invoker.py
```diff
@@ -29,7 +32,7 @@ def set_param(host: str, param_key: str, param_val: str) -> str:
 
 
 def make_error_output(err: Exception) -> Dict[str, str]:
-    """Make a more digestable error output."""
+    """Make a more digestible error output."""
```
zekenautilus, line 35: lol nice

ergo/amqp_invoker.py
```diff
@@ -47,123 +50,90 @@ class AmqpInvoker(Invoker):
     def __init__(self, invocable: FunctionInvocable) -> None:
         super().__init__(invocable)
 
-        host = self._invocable.config.host
-        heartbeat = self._invocable.config.heartbeat
+        self._connection = kombu.Connection(self._invocable.config.host, heartbeat=self._invocable.config.heartbeat)
```
zekenautilus, line 53: maybe update https://github.com/nautiluslabsco/ergo/blob/master/ergo/config.py#L124-L131

to:
```
return int(self._heartbeat) if self._heartbeat else 0
```

since it seems like kombu's default value for the heatbeat parameter is 0:
image

ergo/amqp_invoker.py
```diff
@@ -47,123 +50,90 @@ class AmqpInvoker(Invoker):
...
-        self.exchange_name = self._invocable.config.exchange
-        self.component_queue_name = f"{self._invocable.config.func}".replace("/", ":")
-        if self.component_queue_name.startswith(":"):
-            self.component_queue_name = self.component_queue_name[1:]
-        self.instance_queue_name = f"{self.component_queue_name}:{instance_id()}"
-        self.error_queue_name = f"{self.component_queue_name}:error"
+        component_queue_name = f"{self._invocable.config.func}".replace("/", ":")
+        if component_queue_name.startswith(":"):
+            component_queue_name = component_queue_name[1:]
+        self._component_queue = kombu.Queue(name=component_queue_name, exchange=self._exchange, routing_key=str(SubTopic(self._invocable.config.subtopic)), durable=False)
```
zekenautilus, line 59: is there a reason `self._invocable.config.subtopic` doesn't return something of type `SubTopic`?

zachschubert: we could, but it would require refactoring a few other things that i was planning to get to a bit later

ergo/amqp_invoker.py
zachschubert: good catch. perhaps it would be better to give it a real default value, like rabbitmq's suggested 60s?

zekenautilus: 60s sounds good to me, per https://www.rabbitmq.com/heartbeats.html#heartbeats-timeout

ergo/amqp_invoker.py
```diff
@@ -47,123 +50,90 @@ class AmqpInvoker(Invoker):
...
+        instance_queue_name = f"{component_queue_name}:{instance_id()}"
+        self._instance_queue = kombu.Queue(name=instance_queue_name, exchange=self._exchange, routing_key=str(SubTopic(instance_id())), auto_delete=True)
```
zekenautilus, line 61: with `auto_delete` set to True - does this mean the queue is deleted when there are no longer any consumers?

ergo/amqp_invoker.py
```diff
@@ -47,123 +50,90 @@ class AmqpInvoker(Invoker):
...
+        signal.signal(signal.SIGINT, self._handle_sigterm)
+        with self._connection:
+            conn = self._connection
+            consumer: kombu.Consumer = conn.Consumer(queues=[self._component_queue, self._instance_queue], prefetch_count=PREFETCH_COUNT)
+            consumer.register_callback(self._start_handle_message_thread)
+            consumer.consume()
+            while not self._terminating.is_set():
+                try:
+                    # wait up to 1s for the next message before sending a heartbeat
+                    conn.drain_events(timeout=1)
```
zekenautilus, line 80: looking at the docs for drain_events: https://docs.celeryproject.org/projects/kombu/en/stable/reference/kombu.html#kombu.Connection.drain_events

so the logic here is:
- loop until interrupted
- drain events (wait 1 second)
- after 1 second has passed, check the heartbeat per:
https://docs.celeryproject.org/projects/kombu/en/stable/reference/kombu.html#kombu.Connection.heartbeat_check

  - if we default to 60 seconds, the default rate is `2` which means our tick is `60/2 == 30` seconds. unless 30 seconds has passed from the previous heartbeat tick, nothing happens here, we just loop back to drain_events
  - if 30 seconds passed, send heartbeat
  - if 60 seconds passed, sever connection (hit 3rd except)
  - if we get a message, reset loop?

is that correct?

ergo/amqp_invoker.py
```diff
@@ -47,123 +50,90 @@ class AmqpInvoker(Invoker):
...
-            amqp_message = cast(aio_pika.IncomingMessage, amqp_message)
-            with defer_termination():
-                async with amqp_message.process():
-                    ergo_message = decodes(amqp_message.body.decode('utf-8'))
-                    await self.handle_message(ergo_message, channel_pool)
-
-    async def handle_message(self, message_in: Message, channel_pool: aio_pika.pool.Pool):
+    def _start_handle_message_thread(self, body: str, message: kombu.message.Message):
+        # _handle_sigterm will wait for _handle_message to release this semaphore
+        self._pending_invocations.acquire(blocking=False)
```
zekenautilus, line 93: should this be blocking=True?

ergo/amqp_invoker.py
zachschubert: that's right. if the broker deletes an instance queue because a connection was dropped for too long, then that instance will automatically re-create it [here](https://github.com/nautiluslabsco/ergo/blob/4a87e3930b660e85d7369d401d0265b4048ab8ea/ergo/amqp_invoker.py#L126) once it reconnects. but those IDs are otherwise ephemeral, so we want to make sure the broker deletes those queues after their instances terminate

ergo/amqp_invoker.py
zachschubert: that's correct. also, note that conn.drain_events blocks on callbacks for any events that come in- `_start_handle_message_thread`, in this case. That's why `_start_handle_message_thread` quickly kicks off a task before returning, to ensure that we continue sending heartbeats in case the task takes a long time to finish.

ergo/amqp_invoker.py
```diff
@@ -47,123 +50,90 @@ class AmqpInvoker(Invoker):
...
+            while not self._terminating.is_set():
+                try:
+                    # wait up to 1s for the next message before sending a heartbeat
+                    conn.drain_events(timeout=1)
+                except socket.timeout:
+                    conn.heartbeat_check()
+                except conn.recoverable_connection_errors:
+                    logger.warning("connection closed. reviving.")
+                    conn = self._connection.clone()
+                    conn.ensure_connection()
```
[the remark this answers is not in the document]
zachschubert: found it [here](https://github.com/celery/kombu/blob/master/kombu/connection.py#L384-L438)

ergo/amqp_invoker.py
zachschubert: yeah, good catch

zachschubert: https://github.com/nautiluslabsco/ergo/pull/70/commits/f3772e290e6077c7c61ca88f06b0c3f9b56c356e

zekenautilus:
⚡

PR #71, opened 2022-03-14
Schubert/arg mapping bugfix pd 690

zachschubert:
My previous fix for argument mapping turned out to be insufficient, leading to the behavior reported [here](https://nautiluslabs.atlassian.net/browse/INF-148). See the new tests in test/integration/amqp/test_argument_binding.py` for a more comprehensive description of the behavior that's now definitely correctly supported.


ergo/context.py
```diff
@@ -10,9 +10,9 @@ class Envelope:
     """
     Use this container to pass data to ergo with custom routing parameters.
 
-    >>> hardcoded_subscriber = "interested_party"
+    >>> ad_hoc_topic = "interested_party"
     >>> def my_handler(context: Context):
-    ...     return context.envelope("my_return_val", topic=f"{context.pubtopic}.{hardcoded_subscriber}")
+    ...     return context.envelope("my_return_val", topic=f"{context.pubtopic}.{ad_hoc_topic}")
```
zachschubert, line 15: unrelated nit

ergo/message.py
```diff
@@ -21,14 +18,6 @@ class Message:
...
     traceback: Optional[str] = None
 
-    def get(self, key: str, default=None):
-        value = pydash.get(self.data, key)
-        if value is not None:
-            return value
-        if key == DATA_KEY:
-            return self.data
-        return default
-
```
zachschubert, line 31: i decided to consolidate this logic into FunctionInvoker.assemble_arguments

test/integration/amqp/test_reply_to.py
```diff
@@ -8,7 +8,7 @@
 """
 
 
-def shout(context: Context, message, capitalized=None):
+def shout(context: Context, message=None, capitalized=None):
```
zachschubert, line 11: this should have been failing previously

ergo/function_invocable.py
```diff
@@ -124,6 +122,24 @@ def invoke(self, message_in: Message) -> Generator[Message, None, None]:
...
 
+    def assemble_arguments(self, message: Message, context: Context) -> dict:
+        """
+        assemble_arguments maps a message and a context to a dictionary of keyword arguments to pass to this invoker's
+        handler.
+        """
+        kwargs = {}
+        for param, default in self._params.items():
+            # this is the complete collection of data that this invoker's handler has access to
+            exposed_data = {CONTEXT_KEY: context, DATA_KEY: message.data}
```
zekenautilus, line 133: nit - this can be moved out of the for loop?

zekenautilus, line 133: nit - this can be moved out of the for loop?

ergo/function_invocable.py
```diff
@@ -124,6 +122,24 @@ def invoke(self, message_in: Message) -> Generator[Message, None, None]:
...
+            # ergo's canonical name for this param, which the configuration may have a custom mapping for
+            ergo_param_name = self.config.args.get(param, param)
+            argument = pydash.get(exposed_data, ergo_param_name) or pydash.get(exposed_data, f"{DATA_KEY}.{param}") or default
```
zekenautilus, line 136: this second case `pydash.get(exposed_data, f"{DATA_KEY}.{param}")` is intentional correct?

And it's just stating if we don't have a custom mapping for something, assume this parameter name is nested under `data` and give it to the component?

zekenautilus, line 136: this second case `pydash.get(exposed_data, f"{DATA_KEY}.{param}")` is intentional correct?

And it's just stating if we don't have a custom mapping for something, assume this parameter name is nested under `data` and give it to the component?

zachschubert: > this second case pydash.get(exposed_data, f"{DATA_KEY}.{param}") is intentional correct?  
using `param` there instead of `ergo_param_name` used to be necessary, but (based on other changes) it apparently no longer is. i'll swap it out, because it does _look_ wrong.

> And it's just stating if we don't have a custom mapping for something, assume this parameter name is nested under data and give it to the component?
that's right

test/integration/amqp/test_argument_binding.py
```diff
@@ -0,0 +1,80 @@
...
+
+    args:
+      - my_context: context
+      - my_param: data
+
+    ergo should bind the full payload to `my_param`
+    """
+    results = Queue(routing_key=component.pubtopic)
+    publish(component.subtopic, foo="bar")
+    assert results.consume()["data"] == {"foo": "bar"}
```
zekenautilus, line 28: looking at https://github.com/nautiluslabsco/ergo/blob/master/test/integration/utils/amqp.py to understand these tests. 

is `component` a pass-through component? It just spits out whatever input it gets?

Here we're passing a message body of:
`{"foo": "bar"}` 

and when component operates on it (using ergo in this test PR) it publishes:
`{"data": {"foo": "bar"}}`?

zekenautilus: oh I think I see - `handler_with_mapped_params` is our injected function.

We're mapping `data` to `my_param` via the `args` parameter to the `amqp_component` decorator.

So here what's being tested is that when the invoked function `handler_with_mapped_params` consumes a test message (that's part of the `**message` arg to `publish`) we're verifying that:
1. `foo="bar"` turns into `{"data": {"foo": "bar"}}` on publish
2. `assemble_arguments` then uses the args config of `"my_param": "data"` to translate this payload by key from `data` to the parameter `my_param`
3. `handler_with_mapped_params` is correctly invoked (without exceptions) and publishes the parameter `my_param` (which is now the dictionary `{"foo": "bar"}`

is that correct?

ergo/function_invocable.py
zachschubert: ⚡

zachschubert: ⚡

test/integration/amqp/test_argument_binding.py
```diff
@@ -0,0 +1,80 @@
+from test.integration.utils.amqp import Queue, amqp_component, publish
+
+from ergo.context import Context
+
+"""
+These tests assert that ergo can correctly bind message data to a custom parameter using the `args` configuration attribute.
+"""
+
+
+def handler_with_mapped_params(my_context: Context, my_param):
```
zekenautilus, line 10: what's the behavior like if our function signature was instead:

```
def handler_with_mapped_params(my_context: Context, my_param, infrequent_directive=False):
    retval = {"my_param": my_param}
    if not infrequent_directive:
        retval["zeke is"] = "dumb"
    else:
        # do rare, special thing that we don't always expect in the message payload.
        retval["zeke is"] = "smart"
    return retval
```

if we omitted `infrequent_directive` from our message payload - would a TypeError be raised? For example, would the following happen:
```
@amqp_component(handler_with_mapped_params
publish(component.subtopic, args={"my_param": "data.foo", "my_context": "context", "my_param": "data.boop"})
    ...
    publish(component.subtopic, foo="bar") # omitted case
    actual = results.consume()["data"]
    assert actual["my_param"] == "bar"
    assert actual["zeke is"] == "dumb"
    
    publish(component.subtopic, foo="bar", "boop"=True)
    actual = results.consume()["data"]
    assert actual["my_param"] == "bar"
    assert actual["zeke is"] == "smart"
```

?

test/integration/amqp/test_argument_binding.py
anttimoelsae: Thanks Zeke!

test/integration/amqp/test_argument_binding.py
zachschubert: [These](https://github.com/nautiluslabsco/ergo/blob/faf52322755558d37e01e5ad5a961a686cd8553b/test/integration/amqp/test_amqp.py#L22-L49) tests cover that behavior specifically.

zekenautilus (approved):
🚀

ahuman-bean (approved):

PR #74, opened 2022-03-16
added timestamp to error messages

anttimoelsae:
Context: Currently when looking at messages on an error queue it is very difficult to decipher when they arrived

This PR:
 - adds a timestamp to metadata of message payload for messages hitting error queue
 - adding `"metadata": {"timestamp": "2022-03-16T18:38:20.972894Z"}`

Notes: could additionally show unix timestamp property of RMQ messages if there's no drawback to that


ergo/amqp_invoker.py
```diff
@@ -122,19 +121,21 @@ async def run_queue_loop(self, channel_pool: aio_pika.pool.Pool, queue: aio_pika
     async def handle_message(self, message_in: Message, channel_pool: aio_pika.pool.Pool):
         try:
             async for message_out in self.do_work(message_in):
+                raise Exception('ex')
```
zachschubert, line 124:
```suggestion
```

zachschubert (changes requested):
looks good otherwise

ergo/amqp_invoker.py
anttimoelsae: ouch

jaechoe (approved):
lgtm

zachschubert (approved):

PR #76, opened 2022-03-24
give precedence to namespace files passsed via CLI

zachschubert:
see the new test docstring

### DIFF
--- a/ergo/ergo_cli.py
+++ b/ergo/ergo_cli.py
@@ -1,6 +1,7 @@
 """Summary."""
 import datetime
 import os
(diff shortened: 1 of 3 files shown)


ergo/ergo_cli.py
```diff
@@ -40,14 +40,18 @@ def get_version_path() -> str:
 
 
 def load_config(*config_paths: str) -> Config:
-    merged_data = {}
+    unmerged_data = []
```
ahuman-bean, line 43: `List[dict]`?

PR #77, opened 2022-03-25
pin jinja2 to fix quart import error

zachschubert:
Quart doesn't pin its jinja2 [requirement](https://github.com/pgjones/quart/blob/main/pyproject.toml#L35), but it raises an ImportError with the latest minor version. So for now, we'll have to pin jinja2 ourselves.

image

### DIFF
--- a/ergo/version.py
+++ b/ergo/version.py
@@ -7,7 +7,7 @@ Attributes:
 import subprocess
 import sys
 
-VERSION = '0.8.3-alpha'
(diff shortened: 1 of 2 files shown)


zachschubert:
This was [fixed](https://github.com/pgjones/quart/issues/143) in quart two hours ago, but they haven't made a new release yet.

rickyviank (approved):

PR #81, opened 2022-03-30
fix json transcoding bug

zachschubert:
We have a mismatch in our transcoding functions, where [message.encodes](https://github.com/nautiluslabsco/ergo/blob/3a4242b3999bc4f9d1356c359beec3e0b10418d4/ergo/message.py#L36) uses json.dumps to serialize outbound payloads, while [message.decodes](https://github.com/nautiluslabsco/ergo/blob/3a4242b3999bc4f9d1356c359beec3e0b10418d4/ergo/message.py#L23) uses the cleverer json**s**.loads to deserialize inbound payloads. The latter will automatically decode iso-formatted date strings into datetime objects, which will break any component that depends on receiving dates as strings.

### DIFF
--- a/ergo/amqp_invoker.py
+++ b/ergo/amqp_invoker.py
@@ -85,6 +85,8 @@ class AmqpInvoker(Invoker):
                 except socket.timeout:
                     conn.heartbeat_check()
                 except conn.recoverable_connection_errors:
+                    if self._terminating.is_set():
+                        continue
                     logger.warning("connection closed. reviving.")
                     conn = self._connection.clone()
                     conn.ensure_connection()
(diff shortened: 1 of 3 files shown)


ergo/amqp_invoker.py
```diff
@@ -85,6 +85,8 @@ def start(self) -> int:
                 except socket.timeout:
                     conn.heartbeat_check()
                 except conn.recoverable_connection_errors:
+                    if self._terminating.is_set():
+                        continue
```
zachschubert, line 89: unrelated nit

rickyviank (approved):
bug make sense and thanks for the quick fix @zachschubert !

PR #86, opened 2022-06-28
[Inf 250] feat(queue clean up): unbind component at shutdown

anttimoelsae:
- if component's subtopic gets changed, the old binding is left straggling
 - this change removes bindings from the component's queue at component shutdown -> when a new component instance is spun up it binds back to a 'clean' queue
[INF-250](https://nautiluslabs.atlassian.net/browse/INF-250?atlOrigin=eyJpIjoiY2U4YzU4ZjIzMmMwNDRiMjgwODgzYmRiZTJkMDczYWIiLCJwIjoiaiJ9)

### DIFF
--- a/ergo/amqp_invoker.py
+++ b/ergo/amqp_invoker.py
@@ -61,8 +61,6 @@ class AmqpInvoker(Invoker):
         if component_queue_name.startswith(":"):
             component_queue_name = component_queue_name[1:]
         self._component_queue = kombu.Queue(name=component_queue_name, exchange=self._exchange, routing_key=str(SubTopic(self._invocable.config.subtopic)), durable=False, channel=self._connection.channel())
-        print('bindings', dir(self._component_queue))
-        print('bindings', self._component_queue.as_dict)
         instance_queue_name = f"{component_queue_name}:{instance_id()}"
         self._instance_queue = kombu.Queue(name=instance_queue_name, exchange=self._exchange, routing_key=str(SubTopic(instance_id())), auto_delete=True)
         error_queue_name = f"{component_queue_name}:error"
@@ -82,6 +80,9 @@ class AmqpInvoker(Invoker):
 
             while not self._terminating.is_set():
                 try:
+                    # lazy load binding
+                    # to prevent unbinding all consumers w same subtopic at shutdown
+                    self._component_queue.queue_bind(channel=self._connection.channel())
                     # wait up to 1s for the next message before sending a heartbeat
                     conn.drain_events(timeout=1)
                 except socket.timeout:
@@ -144,9 +145,7 @@ class AmqpInvoker(Invoker):
     def _shutdown(self, signum, *_):
         self._terminating.set()
         self._pending_invocations.acquire(blocking=True, timeout=TERMINATION_GRACE_PERIOD)
-        consumer_count = self._component_queue.queue_declare()[2]
-        if consumer_count <= 1:
-            self._component_queue.queue_unbind()
+        self._component_queue.queue_unbind()
         self._connection.close()
         signal.signal(signum, 0)
         signal.raise_signal(signum)
--- a/ergo/amqp_invoker.py
+++ b/ergo/amqp_invoker.py
@@ -144,7 +144,9 @@ class AmqpInvoker(Invoker):
     def _shutdown(self, signum, *_):
         self._terminating.set()
         self._pending_invocations.acquire(blocking=True, timeout=TERMINATION_GRACE_PERIOD)
-        self._component_queue.queue_unbind()
+        consumer_count = self._component_queue.queue_declare()[2]
+        if consumer_count <= 1:
+            self._component_queue.queue_unbind()
         self._connection.close()
         signal.signal(signum, 0)
         signal.raise_signal(signum)
--- a/ergo/amqp_invoker.py
+++ b/ergo/amqp_invoker.py
@@ -20,6 +20,7 @@ from ergo.topic import PubTopic, SubTopic
 from ergo.util import extract_from_stack, instance_id
 
 import sys
+
 logger = logging.getLogger(__name__)
 
 PREFETCH_COUNT = 1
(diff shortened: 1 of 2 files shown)


ergo/amqp_invoker.py
```diff
@@ -60,12 +61,11 @@ def __init__(self, invocable: FunctionInvocable) -> None:
         component_queue_name = f"{self._invocable.config.func}".replace("/", ":")
         if component_queue_name.startswith(":"):
             component_queue_name = component_queue_name[1:]
-        self._component_queue = kombu.Queue(name=component_queue_name, exchange=self._exchange, routing_key=str(SubTopic(self._invocable.config.subtopic)), durable=False)
+        self._component_queue = kombu.Queue(name=component_queue_name, exchange=self._exchange, routing_key=str(SubTopic(self._invocable.config.subtopic)), durable=False, channel=self._connection.channel())
```
anttimoelsae, line 64: due to `kombu.exceptions.NotBoundError: Can't call method on Queue not bound to a channel`

ergo/amqp_invoker.py
```diff
@@ -142,6 +143,7 @@ def _producer(self) -> kombu.Producer:
     def _shutdown(self, signum, *_):
         self._terminating.set()
         self._pending_invocations.acquire(blocking=True, timeout=TERMINATION_GRACE_PERIOD)
+        self._component_queue.queue_unbind()
```
zekenautilus, line 181: does this handle the case where we might have multiple tasks / consumers subscribing to the same queue?

for example what if we had one (logical) `yeller` that is made by `yeller.py`, has queue `yeller` and subscribes to topic `#.word.#`

but what if that component also has 5 ecs tasks backing it - if 1 ecs task is restarted will it remove the `#.word.#` binding temporarily losing messages to this queue?

zekenautilus:
also don't forget to increment the `version` for this change!

ergo/amqp_invoker.py
zachschubert: i believe we would indeed risk losing messages, and the only way around it would be to iterate over the bindings (there's a `self._component_queue.list_bindings`) at startup time and remove the ones that don't match the current subtopic.

anttimoelsae: fixed with checking for number of consumers at shutdown

zekenautilus: I wonder if you could use `kombu.utils.amq_manager` on startup: https://docs.celeryq.dev/projects/kombu/en/stable/_modules/kombu/utils/amq_manager.html#get_manager

to have access to the amazonmq's management api and then in turn call something like:
https://pyrabbit.readthedocs.io/en/latest/api.html#pyrabbit.api.Client.get_queue_bindings

to get all the queue bindings for the queue, and remove any that don't match the current binding you want to add, by calling something like:

https://pyrabbit.readthedocs.io/en/latest/api.html#pyrabbit.api.Client.delete_binding

cc @zachschubert @anttimoelsae

anttimoelsae: > i believe we would indeed risk losing messages, and the only way around it would be to iterate over the bindings (there's a `self._component_queue.list_bindings`) at startup time and remove the ones that don't match the current subtopic.

I don't see that method on the queue. I am only able to get the binding that is defined when the consumer is started, not all bindings for the queue. Thinking would need to get those from the server as per @zekenautilus suggestion.

anttimoelsae: > I wonder if you could use `kombu.utils.amq_manager` on startup: https://docs.celeryq.dev/projects/kombu/en/stable/_modules/kombu/utils/amq_manager.html#get_manager
> 
> to have access to the amazonmq's management api and then in turn call something like: https://pyrabbit.readthedocs.io/en/latest/api.html#pyrabbit.api.Client.get_queue_bindings
> 
> to get all the queue bindings for the queue, and remove any that don't match the current binding you want to add, by calling something like:
> 
> https://pyrabbit.readthedocs.io/en/latest/api.html#pyrabbit.api.Client.delete_binding
> 
> cc @zachschubert @anttimoelsae

I am (I suppose incorrectly) asssuming that the old task gets killed before the new one is started. In that case the check on number of consumers alone would not work.

Pyrabbit is almost older than I am, so it does not work out of the box, but I could just build a mini-client based on the [RMQ api](https://b-565373c7-7456-4c97-a192-841339caa9ce.mq.us-east-1.amazonaws.com/api/index.html). To get the bindings. Unsure if it is wise to have core infra code depending on http requests like that. I would need @zachschubert input on that. Or if I am just missing sth w/kombu not able to access the bindings.

herrmt:
This feels like a really clean way to perform the cleanup requested in the ticket.

One question for a sanity check @anttimoelsae  - we unbind if the consumer count is 1 because that means the component shutting down is the ONLY active consumer and therefore consumers will go to zero when we shut down, correct?

anttimoelsae:
> This feels like a really clean way to perform the cleanup requested in the ticket.
> 
> One question for a sanity check @anttimoelsae - we unbind if the consumer count is 1 because that means the component shutting down is the ONLY active consumer and therefore consumers will go to zero when we shut down, correct?

Correct. However, I am now unsure if zeke's scenario of starting and binding a new component before the old one shuts down is something we need to worry about. And if it is, the only way to solve it is to fetch a list of bindings via a http request from the AWS MQ server, which I am not sure is wise to include in the ergo code. I guess the safe way to do it is to shut down the component if the request fails after some retries.

ergo/amqp_invoker.py
```diff
@@ -78,8 +77,12 @@ def start(self) -> int:
             consumer: kombu.Consumer = conn.Consumer(queues=[self._component_queue, self._instance_queue], prefetch_count=PREFETCH_COUNT, accept=["json"])
             consumer.register_callback(self._start_handle_message_thread)
             consumer.consume()
+
             while not self._terminating.is_set():
                 try:
+                    # lazy load binding
+                    # to prevent unbinding all consumers w same subtopic at shutdown
+                    self._component_queue.queue_bind(channel=self._connection.channel())
```
herrmt, line 85: Oh I like this - much less fussy than initializing an entire queue and reliant on all the same connections.

PR #87, opened 2022-07-19
[VOTR-119] fix(falsey argument binding): bind falsey arguments to handler params

zachschubert:
There’s a bug in FunctionInvocable.assemble_arguments that causes it to fail to bind any data that equals False (e.g. 0) to a handler param.


zekenautilus (approved):
👍

PR #88, opened 2022-07-27
generate instance id from current time

zachschubert:
This generates the ergo instance ID (and instance queue name) from the current time instead of a random string. This is moderately helpful (for me, right now) for development, by allowing amqp console users to distinguish between component instances and route messages accordingly.


zekenautilus (approved):
👍

PR #89, opened 2022-08-01
fix(ergo graph): detect more ergo configs; fixed overlapping labels by allowing nodes to resize themselves; put topics in boxes

xytxytxyt:
currently `ergo graph` doesn't pick up all ergo configs, and it also has a problem with overflowing label text (screenshot because it generates a pdf):
image

after this change:
ergo gv

### DIFF
--- a/ergo/schematic.py
+++ b/ergo/schematic.py
@@ -53,10 +53,13 @@ def load_configs(folders: List[str]) -> List[Dict[str, Union[None, str, List[str
     for folder in folders:
         yaml_files = glob.glob(os.path.join(folder, '**/*.y*ml'), recursive=True)
         for yaml_file in yaml_files:
-            if not os.path.basename(yaml_file).startswith('serverless'):
+            if not os.path.basename(yaml_file).startswith('serverless'):  # skip serverless config files
                 with open(yaml_file, 'r', encoding='utf8') as stream:
                     config = yaml.safe_load(stream)
                     if config and 'func' in config:
+                        # example:
+                        # root_folder/api_connector/ship_dc/api_connector.yaml becomes
+                        # api_connector/ship_dc/api_connector
                         component_name = yaml_file[len(os.path.normpath(folder)) + 1:].split('.')[0]
                         config['name'] = component_name
                         configs.append(config)
--- a/ergo/schematic.py
+++ b/ergo/schematic.py
@@ -75,9 +75,9 @@ def topics(dot: graphviz.Digraph, configs: List[Dict[str, Union[None, str, List[
     for config in configs:
         for topic_element in format_topic('pubtopic', config):
             dot.edge(format_component(config)[0], topic_element[0])
-            dot.node(*topic_element)
+            dot.node(*topic_element, shape='box')
         for topic_element in format_topic('subtopic', config):
-            dot.node(*topic_element)
+            dot.node(*topic_element, shape='box')
             dot.edge(topic_element[0], format_component(config)[0])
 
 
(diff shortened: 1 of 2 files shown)


ergo/schematic.py
```diff
@@ -50,14 +51,15 @@ def load_configs(folders: List[str]) -> List[Dict[str, Union[None, str, List[str
     configs = []
 
     for folder in folders:
-        for name in list_paths(folder):
-            try:
-                with open(f'{folder}/{name}/{name}.yaml', 'r', encoding='utf8') as stream:
+        yaml_files = glob.glob(os.path.join(folder, '**/*.y*ml'), recursive=True)
+        for yaml_file in yaml_files:
+            if not os.path.basename(yaml_file).startswith('serverless'):
```
zekenautilus, line 56: might be worth leaving a comment here. To make it clear we're avoiding a config file for another tool (serverless).

zekenautilus:
@xytxytxyt it looks like for some of the components the folder names are repeated - is this intentional? 

image

ergo/schematic.py
```diff
@@ -50,14 +51,15 @@ def load_configs(folders: List[str]) -> List[Dict[str, Union[None, str, List[str
...
+                with open(yaml_file, 'r', encoding='utf8') as stream:
                     config = yaml.safe_load(stream)
-                    config['name'] = name
-                    configs.append(config)
-            except FileNotFoundError:
-                pass
+                    if config and 'func' in config:
+                        component_name = yaml_file[len(os.path.normpath(folder)) + 1:].split('.')[0]
```
zekenautilus, line 60: what exactly is this indexing into?

for `component_name` in the circleci stuff for QED I sorta went the route of just reversing directory ordering and connecting the folders with underscores.

`api_connector/kyma` -> `kyma_api_connector`

which could maybe be done here? https://github.com/nautiluslabsco/nl-circle-orb/blob/main/src/scripts/autogen.py#L95-L102

though tbf, I feel like directory path is pretty clear for a given repository.

ergo/schematic.py
```diff
@@ -73,9 +75,9 @@ def topics(dot: graphviz.Digraph, configs: List[Dict[str, Union[None, str, List[
     for config in configs:
         for topic_element in format_topic('pubtopic', config):
             dot.edge(format_component(config)[0], topic_element[0])
-            dot.node(*topic_element)
+            dot.node(*topic_element, shape='box')
```
zekenautilus, line 78: image
 👀

ergo/schematic.py
```diff
@@ -104,7 +106,7 @@ def components(dot: graphviz.Digraph, configs: List[Dict[str, Union[None, str, L
         dot (graphviz.Digraph): Description
         configs (List[Dict[str, Union[None, str, List[str]]]]): Description
     """
-    dot.attr('node', shape='circle', width='1', color='#ffffff80', penwidth='2', fixedsize='true')
+    dot.attr('node', shape='circle', width='1', color='#ffffff80', penwidth='2')
```
zekenautilus, line 109: 👍

zekenautilus (approved):
👍 might want to upgrade the PATCH part of the current version?

Most of the comments I left aren't super important, except for the folder duplication I noticed!

ergo/schematic.py
xytxytxyt: i take the path from the root folder (which is passed in as a parameter) to the config file as the component name; for example this component: https://github.com/nautiluslabsco/qed/blob/master/components/api_connector/ship_dc/api_connector.yaml would be named `api_connector/ship_dc/api_connector`

so here, `yaml_file` would be `root_folder/api_connector/ship_dc/api_connector.yaml` so this line just removes the root folder and the yaml extension

it indexes into `yaml_file` with the length of the root folder, + 1 for the path separator... the `normpath()` removes duplicate separators etc

PR #90, opened 2022-08-02
ack early

zachschubert:
Long running components (routes & neptune) need to acknowledge amqp message receipt before they finish, or else they may be disconnected by the broker. As far as I can tell this is the amqp `best practice`, and that components that need to defer acking until they're done should require a flag (as per celery's `acks_late`). 

This also bumps ergo to a non-alpha version. In my experience, requiring components to pin to a patch version of ergo or else install it with an `allow-prereleases` flag has been a hassle with no upside. But if there's a reason to keep this in alpha then I can go back to it.

### DIFF
--- a/ergo/amqp_invoker.py
+++ b/ergo/amqp_invoker.py
@@ -107,6 +107,8 @@ class AmqpInvoker(Invoker):
         # sequentially to guarantee that messages are acknowledged in the order they're received
         with self._handler_lock:
             try:
+                # we acknowledge message receipt eagerly, in order to support long running tasks that
+                # may exceed their broker's `consumer_timeout` setting (30m by default)
                 ack()
                 ergo_message = decodes(body)
                 self._handle_message_inner(ergo_message)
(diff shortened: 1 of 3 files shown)


ergo/amqp_invoker.py
```diff
@@ -107,10 +107,10 @@ def _handle_message(self, body: str, ack: Callable):
         # sequentially to guarantee that messages are acknowledged in the order they're received
         with self._handler_lock:
             try:
+                ack()
```
ahuman-bean, line 110: @zachschubert do you mind leaving a summary of what you put in the PR description here? i think it might be useful for readers to know

ahuman-bean (approved):

herrmt (approved):

ergo/amqp_invoker.py
```diff
@@ -107,10 +107,13 @@ def _handle_message(self, body: str, ack: Callable):
...
         with self._handler_lock:
             try:
+                if self._invocable.config.acks_early:
+                    ack()
                 ergo_message = decodes(body)
                 self._handle_message_inner(ergo_message)
             finally:
-                ack()
+                if not self._invocable.config.acks_early:
+                    ack()
```
ahuman-bean, line 116: is there a way to elide this branch? i.e. map the flag to the handler

PR #94, opened 2022-08-09
[VOTR-185] autodelete component queues

zachschubert:
This PR reverts the changes in https://github.com/nautiluslabsco/ergo/pull/86, which don't seem to be working as intended. That change was recently installed by the routes-generator staging consumers, which have since begun consistently falling into a [connection error loop](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fecs$252Froutes-generator-staging/log-events/ecs$252Froutes$252F626d49602bc443909c2083497c6fe8f8). Meanwhile, the broker is reporting hundreds of consumer bindings on their [component queue](https://b-4578b105-1d2d-4ecf-8f90-63ff45aa4fbc.mq.us-east-1.amazonaws.com/#/queues/%2F/routes%3Ahandler.py%3Agenerate_routes). I don't have a theory for the reverted change could be failing, but I don't see any other commits that could account for it. 

I also added the `auto-delete` flag to the component queue declaration, which should have the same effect.

### DIFF
--- a/ergo/amqp_invoker.py
+++ b/ergo/amqp_invoker.py
@@ -61,11 +61,12 @@ class AmqpInvoker(Invoker):
         component_queue_name = f"{self._invocable.config.func}".replace("/", ":")
         if component_queue_name.startswith(":"):
             component_queue_name = component_queue_name[1:]
-        self._component_queue = kombu.Queue(name=component_queue_name, exchange=self._exchange, routing_key=str(SubTopic(self._invocable.config.subtopic)), durable=False, channel=self._connection.channel())
+        self._component_queue = kombu.Queue(name=component_queue_name, exchange=self._exchange, routing_key=str(SubTopic(self._invocable.config.subtopic)), durable=False)
         instance_queue_name = f"{component_queue_name}:{instance_id()}"
         self._instance_queue = kombu.Queue(name=instance_queue_name, exchange=self._exchange, routing_key=str(SubTopic(instance_id())), auto_delete=True)
         error_queue_name = f"{component_queue_name}:error"
         self._error_queue = kombu.Queue(name=error_queue_name, exchange=self._exchange, routing_key=error_queue_name, durable=False)
+
         self._terminating = threading.Event()
         self._pending_invocations = threading.Semaphore()
         self._handler_lock = threading.Lock()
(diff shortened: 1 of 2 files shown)


ahuman-bean (approved):

anttimoelsae:
> This PR reverts the changes in #86, which don't seem to be working as intended. That change was recently installed by the routes-generator staging consumers, which have since begun consistently falling into a [connection error loop](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fecs$252Froutes-generator-staging/log-events/ecs$252Froutes$252F626d49602bc443909c2083497c6fe8f8). Meanwhile, the broker is reporting hundreds of consumer bindings on their [component queue](https://b-4578b105-1d2d-4ecf-8f90-63ff45aa4fbc.mq.us-east-1.amazonaws.com/#/queues/%2F/routes%3Ahandler.py%3Agenerate_routes). I don't have a theory for the reverted change could be failing, but I don't see any other commits that could account for it.
> 
> I also added the `auto-delete` flag to the component queue declaration, which should have the same effect.

You saw this only in staging, not in QA? I suppose in that case there is some unpredictable behavior and we have to scrap this strategy.

The goal was to remove "straggler" bindings in the case that the subtopic of the component is changed. I.e. we deploy a new version, a new binding is added to the queue, but the binding from the previous deployment does not get removed. In this case the queue would never reach 0 consumers, which is when auto-delete would delete the queue.

zachschubert:
> You saw this only in staging, not in QA? I suppose in that case there is some unpredictable behavior and we have to scrap this strategy.

That's right, I never saw it in QA. It's hard to guess why ¯\\_(ツ)_/¯

> The goal was to remove "straggler" bindings in the case that the subtopic of the component is changed. I.e. we deploy a new version, a new binding is added to the queue, but the binding from the previous deployment does not get removed. In this case the queue would never reach 0 consumers, which is when auto-delete would delete the queue.

The broker will also autodelete the queue when all consumers have closed their [connections](https://www.rabbitmq.com/queues.html#temporary-queues), which would happen after a successful deploy that changes that subtopic. I confirmed that this is the case by testing manually.

PR #95, opened 2022-08-09
replace signal.raise_signal with os.kill

zachschubert:


ergo/amqp_invoker.py
```diff
@@ -151,4 +152,4 @@ def _shutdown(self, signum, *_):
         self._component_queue.queue_unbind()
         self._connection.close()
         signal.signal(signum, 0)
-        signal.raise_signal(signum)
```
zachschubert, line 154: `signal.raise_signal` was added in python 3.8

PR #96, opened 2022-08-09
[VOTR-185] fix(tests): fix test flakiness

zachschubert:
I'm seeing two potential causes for test flakiness:
1. The test runner is trying to gracefully shutdown ergo subprocs, which tends to timeout for a particular test (`test_fibonacci`) that generates a lot of amqp messages. But there's no reason to be graceful here, since we're tearing everything down anyways.
2. Ergo's amqp consumer shutdown sequence uses a function that was added in python 3.8, which periodically breaks the 3.7 test runner. I replaced it with something 3.7 compatible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,7 @@ jobs:
           name: run tests
           command: |
             . venv/bin/activate
+            ulimit -n 1024
             make test
             codecov
       - run:
@@ -85,6 +86,7 @@ jobs:
           name: run tests
           command: |
             . venv/bin/activate
+            ulimit -n 1024
             make test
             codecov
       - store_artifacts:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -148,23 +148,15 @@ workflows:
       - test-python-install:
           name: test-python-install-3.7
           version: "3.7"
-          requires:
-            - build_verify_lint
```
zachschubert, line 152: it doesn't seem necessary to make the test runners block on the linter. but i can revert this if anybody objects

xytxytxyt: does it skip the test if its requirement fails?

maybe the intent of this is to say that if linting fails, no need to run the python tests?

zachschubert: yeah that probably is the intent. but i would expect the linter to fail pretty rarely, so having the tests block on it feels expensive. also in the case where i push lint and bugs at the same time, it would be nice to know about them both.

xytxytxyt: makes sense... 

as a side note would it make sense to move the linting commands to the Makefile so we can run them before we push/open a pr?

xytxytxyt (approved):
so far the tests for this pr have passed on the first run, which is a good sign

.circleci/config.yml
zachschubert: yeah sure, i'll do that in a followup

PR #97, opened 2022-08-10
Linter changes

zachschubert:
1. Split `lint` into two make targets:
a. `make lint` emits lint/style errors without fixing anything in-place. This target is run automatically by circle. 
b. `make fix` in-place fixes any errors it can. This target is meant to be run manually by us.
2. Refactor both scripts so that by default they apply only to the python files that have been modified since branching from master. 
3. Remove flake8 and pylint. These linters are arguably redundant with black, except that they can't fix any of the errors they generate.


.circleci/config.yml
```diff
@@ -11,53 +14,24 @@ release_only: &release_only
       only: /[0-9]+(\.[0-9]+)*/
     branches:
       ignore: /.*/
-
 jobs:
-  build_verify_lint:
+  lint:
     docker:
-      - image: circleci/python:3.7
+      - image: cimg/python:3.7
```
zekenautilus, line 20: nice!

.circleci/config.yml
```diff
@@ -11,53 +14,24 @@ release_only: &release_only
...
-            . venv/bin/activate
-            make dev
-      - save_cache:
-          key: v1-dependency-cache-{{ checksum "setup.py" }}-{{ checksum "Makefile" }}
-          paths:
-            - "venv"
+      - python/install-packages:
+          pkg-manager: pip
+          pip-dependency-file: dev-requirements.txt
+          cache-version: v1
```
zekenautilus, line 29: nice!

scripts/fix.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+source ./scripts/shared.sh
+
+mod=${1:-$(diffed_files)}
```
zekenautilus, line 5: what's the reasoning behind only applying linting to diffed files?

zekenautilus (approved):
🔥 big improvement!

scripts/fix.sh
zachschubert: otherwise the first person who touches ergo after we merge this is gonna be responsible for fixing 3985238547932 bits of old lint 😄

.circleci/config.yml
```diff
@@ -143,21 +117,22 @@ jobs:
...
+      - test:
+          name: test-python-3.10
           version: "3.10"
       - deploy:
           <<: *release_only
           requires:
-            - build_verify_lint
\ No newline at end of file
+            - lint
+            - test-python-3.7
```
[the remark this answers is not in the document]
zachschubert: the prior `build_verify_lint` job did linting and testing in a python 3.7 environment. these jobs do the same, but in parallel and with (imo) greater clarity

mypy.ini
```diff
@@ -17,7 +17,3 @@ disallow_any_unimported = True
 follow_imports = normal
 disallow_any_expr = False
 disallow_any_decorated = True
-disallow_any_explicit = True
```
xytxytxyt, line 20: what's the reasoning behind this change?

mypy.ini
```diff
@@ -17,7 +17,3 @@ disallow_any_unimported = True
...
-
-[mypy-src.types]
```
xytxytxyt, line 22: out of curiosity because i've never used mypy, what was this section for?

.circleci/config.yml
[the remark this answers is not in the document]
zekenautilus: I guess what i'm asking is why not have:
```
requires:
- lint
- test-python-3.7
- ...
- test-python-3.10
```
?

xytxytxyt (approved):
pylint and flake8 is more redundant with autopep8 than with black

zeke knows how i feel about black lol but oh well... it seems to be a popular tool

in any case neither pylint nor flake8 nor the deleted lint script was in use so they're safe to delete

mypy.ini
zekenautilus: https://mypy.readthedocs.io/en/stable/config_file.html

`mypy` lets you configure cli options via a `mypy.ini` file.

I believe this option:
```
[mypy]
...
disallow_any_explicit = True
```

basically made it s.t. whenever `mypy` was run it would disallow the use of explicit `Any` in the codebase.

so if you had something like:
```
def my_func(my_obj: Any) -> Dict:
    if isinstance(my_obj, str):
        return {"obj_value": f"my{my_obj}"}
``` 

`mypy` would fail and ask you to remove the `Any`

zachschubert: That rule made [this](https://github.com/nautiluslabsco/ergo/blob/b28d4b6dc94597ed2a486a45b091d250346e79b4/ergo/message.py#L14) type annotation illegal. But Message.data literally can be anything, and an explicit annotation is required there by Python.

xytxytxyt: i see... i guess i'd rather have some ignore-this-line directive for certain lines (does mypy have this?) rather than disable the rule

because otherwise it would be too easy to defeat the purpose of mypy by using Any excessively 

but no strong feelings either way

zekenautilus: you can do that as well: https://github.com/nautiluslabsco/ergo/blob/master/ergo/function_invocable.py#L37

by putting:
`# type: ignore` directives on the offending line.

I don't have strong opinions on keeping the `disallow_any_explicit` option vs removing it.

zachschubert: in my experience mypy's non-default settings tend to be impractical. For example, with the mypy.ini we have in master we would have to e.g.

```
ergo/amqp_invoker.py:78: error: Type of variable becomes "Any" due to an unfollowed import  [no-any-unimported]
ergo/amqp_invoker.py:101: error: Argument 3 to "_start_handle_message_thread" becomes "Any" due to an unfollowed import  [no-any-unimported]
ergo/amqp_invoker.py:145: error: Return type becomes "Any" due to an unfollowed import  [no-any-unimported]
```
suppress an error every time we assign the output of a kombu function to a variable, because there are no stubs for that library.
 
```
ergo/amqp_invoker.py:145: error: Type of decorated function contains type "Any" ("Callable[[AmqpInvoker], _GeneratorContextManager[Any]]")  [misc]
```
suppress an error every time we decorate with @contextmanager.

imo running an extra strict mypy on a codebase that's filled with `# type: ignore` feels more like chaos than order

mypy.ini
zachschubert: that sets override parameters for `ergo/src/types.py`, which we renamed to `ergo/ergo/types.py` a long time ago

.circleci/config.yml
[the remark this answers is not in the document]
zachschubert: we could, but it feels excessive to me. we already require branches to be up to date before merging, so hypothetically any of these checks are going to be redundant post-merge

PR #102, opened 2022-10-18
feat: added option to specify error pubtopic

xytxytxyt:
currently in ergo it is not possible for a component to subscribe to the error queues of multiple components

this feature is useful for a 3rd party outage detector that we're working on that will subscribe to errors from api connectors, categorize them, and report them to slack

this pr adds this feature

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -12,3 +12,4 @@ requests
 docker
 pytest
 pytest-timeout
+importlib-metadata==4.13.0
--- a/ergo/schematic.py
+++ b/ergo/schematic.py
@@ -1,5 +1,6 @@
 """Summary."""
 import glob
+import itertools
 import os
 import sys
 from typing import Dict, Generator, List, Tuple, Union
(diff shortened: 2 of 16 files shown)


ergo/schematic.py
```diff
@@ -80,6 +81,9 @@ def topics(dot: graphviz.Digraph, configs: List[Dict[str, Union[None, str, List[
         for topic_element in format_topic('pubtopic', config):
             dot.edge(format_component(config)[0], topic_element[0])
             dot.node(*topic_element, shape='box')
+        for topic_element in format_topic('error_pubtopic', config):
+            dot.edge(format_component(config)[0], topic_element[0])
+            dot.node(*topic_element, shape='octagon')
```
xytxytxyt, line 86: chose octagon because it is used for stop signs: http://www.trafficsign.us/signshape.html

sample: 
image

dev-requirements.txt
```diff
@@ -12,3 +12,4 @@ requests
 docker
 pytest
 pytest-timeout
+importlib-metadata==4.13.0
```
xytxytxyt, line 15: https://exerror.com/entrypoints-object-has-no-attribute-get/

zachschubert (approved):

PR #104, opened 2022-11-04
chore: added python 3.11 to list of tested versions; linting changes

xytxytxyt:
since we want to move to python 3.11, i thought i'd start by running tests in 3.11

circleci/python has been deprecated https://circleci.com/docs/next-gen-migration-guide/ and more importantly they never made a circleci/python:3.11 so i switched to cimg/ and had to update the cache keys and make a few other small changes

i tried to make the linting pass and got it down to only mypy failures... but those are too many and it seems like the whole repo has type errors? in any case that is a project for another day

not updating version because no functional changes

### DIFF
--- a/test/integration/utils/amqp.py
+++ b/test/integration/utils/amqp.py
@@ -34,13 +34,7 @@ class AMQPComponent(FunctionComponent):
     protocol = "amqp"
     instances: List[AMQPComponent] = []
 
-    def __init__(
-        self,
-        func: Callable,
-        subtopic: Optional[str] = None,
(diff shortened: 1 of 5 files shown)


test/integration/utils/amqp.py
```diff
@@ -11,7 +11,7 @@
 import kombu
 import kombu.pools
 import kombu.simple
-from amqp import Channel
+from amqp.channel import Channel
```
xytxytxyt, line 14: this avoids 
image

xytxytxyt:
closing because not a priority

PR #107, opened 2022-12-08
fix: changed autodelete to auto_delete

xytxytxyt:
the correct spelling seems to be auto_delete

sources:
https://docs.celeryq.dev/projects/kombu/en/stable/reference/kombu.html#queue
https://github.com/nautiluslabsco/ergo/blob/b7805bcf1e411df461197432783eae8ad5cb37aa/ergo/amqp_invoker.py#L68


rickyviank (approved):
change looks simple enough! Thanks for including the kombu documentation for cross reference :D 

Tho there's a linting error in CI 👀

zachschubert (approved):

xytxytxyt:
yeah i think the linting has never passed 

(they are mypy errors; the nature of mypy is you can't only check files that changed, and it would need significant effort to fix the whole codebase)

PR #108, opened 2023-06-22
chore: don't auto delete component queue

xytxytxyt:
in https://github.com/nautiluslabsco/ergo/pull/107 i fixed a typo and set component queues to auto delete 

but after using that a bit, i have changed my mind; reasoning below

firstly, according to [the documentation](https://www.rabbitmq.com/queues.html#:~:text=An%20auto%2Ddelete%20queue%20will,TCP%20connection%20with%20the%20server), if so configured, these queues will be deleted when the consumer disconnects... i could not find anything on the internet that says the queue has to be empty... so it deletes the queue even if it's not empty, as long as the consumer has disconnected

also it's difficult to update an existing component to ergo 0.11.3 because:
* queue already exists
* we upgrade to 0.11.3 and deploy
* we try to recreate the queue with auto delete
* this fails because there is already an existing queue with the same name, but without auto delete
* we have to manually delete the queue and start the component, which will recreate the queue with auto delete

on the other hand, if this pr is approved, going to 0.11.4 will be easy because 0.11.3 queues will be auto deleted when the component shuts down for deploy, so there will be no problem recreating them without auto delete

instance queues remain auto delete, which i think is correct: https://github.com/nautiluslabsco/ergo/blob/34644265371a0d12b0292dd65cd0b893b423e626/ergo/amqp_invoker.py#L68

PR #109, opened 2023-06-23
chore: try to resolve current test failures by running with sudo

xytxytxyt:
in https://github.com/nautiluslabsco/ergo/pull/108 unrelated tests which passed last time are now failing

digging deeper, we see failures that look like
```
  File "/home/circleci/project/venv/lib/python3.10/site-packages/hypercorn/config.py", line 261, in _create_sockets
    sock.bind(binding)
PermissionError: [Errno 13] Permission denied
```

it is trying to use localhost port 80, so maybe there was a policy change in circleci to no longer allow that... this pr runs the test with sudo to get around that

references:
https://stackoverflow.com/questions/8633461/how-to-keep-environment-variables-when-using-sudo
https://askubuntu.com/questions/234758/how-to-use-a-python-virtualenv-with-sudo
https://stackoverflow.com/questions/28890634/how-to-get-a-shell-environment-variable-in-a-makefile

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
           name: run tests
           command: |
             . venv/bin/activate
-            make test
+            sudo -E make test
             codecov
       - store_artifacts:
           path: htmlcov/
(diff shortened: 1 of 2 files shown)


herrmt (approved):
Confirmed with Xian this still runs locally for tests. Approving.

